PR #43, opened 2015-03-13
setup.py: Change url to github

msabramo:
The GitHub URL is more useful than the PyPI URL, because this data is most often seen on PyPI.


tartley:
brill, thanks.

Issue #48, opened 2015-04-15
Escape sequences are unprocessed on Windows with TERM set


msiemens:
On Windows, when the environment variable `TERM` is set to any value, escape sequences will be printed unprocessed (no colored output). Concerning the values of `TERM`/the terminals where this happens, I'm still waiting for user feedback (see msiemens/PyGitUp#24).

According to `git bisect` this behaviour has been introduced in 961428e6c1c3a6855da5dedccfc418f295b54c25.

krist10an:
I'm having the same issue using Colorama. I'm using bash from https://msysgit.github.io/, and that has TERM set to "cygwin" by default. Let me know if there is anything special I could test to help you verify (it wasn't clear to me in the linked issue)

wiggin15:
Hmm... Bash should be able to process the escape sequences, because it's supposed to emulate a unix shell. Are you running bash from 'cmd' or from a Cygwin terminal?

It looks like a lot of people are encountering this problem so I think the commit (linked above) should be reverted. I just don't know what to do with Cygwin instead.

krist10an:
I'm using bash that is included in the git package, and it is not started from cmd, but though the start menu. The shortcut command in the start menu is starting a separate command: ```"C:\Program Files (x86)\Git\bin\sh.exe" --login -i```

Also I don't have cygwin installed (and don't want that in the future either :smile:  )

krist10an:
For reference, if I first run ```unset TERM``` then output ANSI with colorama the colors work as expected.

ezk84:
The following script:

```python
#!/bin/env python
from colorama import init, Fore, Back, Style, deinit

init()
print("after init(): " +Fore.RED + 'some red text' +
      Fore.RESET + Back.RESET + Style.RESET_ALL)
print('back to normal now')

deinit()
init(convert=True)
print("after init(convert=True): " + Fore.RED + 'some red text' +
      Fore.RESET + Back.RESET + Style.RESET_ALL)
print('back to normal now')
```

Will output unprocessed escape sequences in the first instance and proper color in the last under Git for Windows. Interestingly, if I pipe the output to `cat`, the color is shown correctly in the first instance, and no colors nor escape characters are shown for the second. 
Under `cmd` the script works as expected.

gvanem:
I had the same issue if e.g. `TERM=VT100` (that's needed in some Unix centric programs here).
I just added:
```
on_emulated_windows = False
```
[...]

PR #50, opened 2015-04-17
Fix #49

### LINKED ISSUE #49: ANSI reset code is printed on Windows on program exit
Given the following Python file:

```python
import colorama
colorama.init()
```

Running it results in this:

```cmd
C:\Users\markus\tmp>python test.py
←[0m
C:\Users\markus\tmp>
```

This is the result of running [`initialise.reset_all`](https://github.com/tartley/colorama/blob/dc06ba4bcc6c641f2f1b685b2207047e6db594f7/colorama/initialise.py#L17) when exiting the script. The `AnsiToWin32` instance that is created isn't instructed to convert ANSI codes, so it just prints the ANSI reset code.

msiemens:
This is a propsed fix for #49 which I've reported. This solution looks somewhat hacky to me as it changes the signature of `initialise.reset_all` and I don't know whether that's okay or not.

### DIFF
--- a/colorama/initialise.py
+++ b/colorama/initialise.py
@@ -14,8 +14,12 @@ wrapped_stderr = sys.stderr
 atexit_done = False
 
 
-def reset_all():
-    AnsiToWin32(orig_stdout).reset_all()
+def reset_all(autoreset, convert, strip):
+    def callback():
+        AnsiToWin32(orig_stdout,
+                    convert=convert,
+                    strip=strip,
+                    autoreset=autoreset).reset_all()
 
 
 def init(autoreset=False, convert=None, strip=None, wrap=True):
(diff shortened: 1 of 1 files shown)


wiggin15:
This seems to be related to your issue, #48. It doesn't happen on environments without TERM set...

I see that TERM causes problems for some users, and I'm trying to think if there's a way to change this. It's necessary to make colorama work on terminal emulators under Windows. Why do you have TERM set on cmd?

wiggin15:
The usage of 'register' here isn't right. Instead of registering the function, you are calling reset_all in initialize and passing the result of the call to 'register'.

msiemens:
As noted in #49, this patch is propably wrong.

Issue #57, opened 2015-05-10
Color Style not reset after RESET when using LIGHT_EX colors in Windows


andyneff:
When printing a Fore.LIGHT*_EX color and attempting to use Fore.RESET, in windows the colors stay "light" instead of returning to their normal state. A RESET_ALL or Style Change is needed to manually fix it.

In Linux, it behaves as expected.
![linux](https://cloud.githubusercontent.com/assets/7596961/7553202/d530ba42-f6b6-11e4-88c5-c631c89e7b06.png)

But in Windows 7, it does not
![windows](https://cloud.githubusercontent.com/assets/7596961/7553203/e107df8a-f6b6-11e4-821c-00ee6f90a992.png)
It's a little hard to see, but on the first line the d and e are bright, When only the d should be bright, and on the second line, the b, c, d, and e are all bright, when again, only the d should be bright.

viswanathct:
Have stumbled across the same issue too. The following code isn't working as expected.
```python
import colorama
from colorama import Fore

Colors = [color for color in dir(Fore) if color[0] != '_']
Colors.remove('RESET')

colorama.init()

for color in sorted(Colors):
	print '%s%s%s' % (getattr(Fore, color), color, Fore.RESET)
	
print '-------------'

for color in Colors:
	print '%s%s%s' % (getattr(Fore, color), color, Fore.RESET)
```

The **BLACK** in the second list is printed in **LIGHTBLACK_EX**.

andyneff:
I seem to have inadvertently created two discussions on here (#57 and #58 )

PR #58, opened 2015-05-10
Fixed bug with Fore.RESET after any LIGHT_EX for Issue #57

andyneff:
Returned the "light" brightness to normal for non-light colors

### DIFF
--- a/colorama/winterm.py
+++ b/colorama/winterm.py
@@ -27,9 +27,10 @@ class WinTerm(object):
         self._default_fore = self._fore
         self._default_back = self._back
         self._default_style = self._style
+        self._light = 0
 
     def get_attrs(self):
-        return self._fore + self._back * 16 + self._style
+        return self._fore + self._back * 16 + (self._style | self._light)
 
     def set_attrs(self, value):
         self._fore = value & 7
@@ -45,9 +46,9 @@ class WinTerm(object):
             fore = self._default_fore
         self._fore = fore
         if light:
-            self._style |= WinStyle.BRIGHT
+            self._light |= WinStyle.BRIGHT
         else:
-            self._style &= ~WinStyle.BRIGHT
+            self._light &= ~WinStyle.BRIGHT
         self.set_console(on_stderr=on_stderr)
 
     def back(self, back=None, light=False, on_stderr=False):
@@ -55,9 +56,9 @@ class WinTerm(object):
             back = self._default_back
         self._back = back
         if light:
-            self._style |= WinStyle.BRIGHT_BACKGROUND
+            self._light |= WinStyle.BRIGHT_BACKGROUND
         else:
-            self._style &= ~WinStyle.BRIGHT
+            self._light &= ~WinStyle.BRIGHT_BACKGROUND
         self.set_console(on_stderr=on_stderr)
 
     def style(self, style=None, on_stderr=False):
--- a/colorama/winterm.py
+++ b/colorama/winterm.py
@@ -46,6 +46,8 @@ class WinTerm(object):
         self._fore = fore
         if light:
             self._style |= WinStyle.BRIGHT
+        else:
+            self._style &= ~WinStyle.BRIGHT
         self.set_console(on_stderr=on_stderr)
 
     def back(self, back=None, light=False, on_stderr=False):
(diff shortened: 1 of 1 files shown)


wiggin15:
Consider this code:

    print colorama.Fore.RED + colorama.Style.BRIGHT + "this should be bright" + colorama.Fore.RESET
    print colorama.Fore.RED + "this should still be bright" + colorama.Fore.RESET

In this case the second print should also be light (the style was not reset), but with your commit it will auto-reset.

Perhaps a different approach would be to remember `self._style` in `fore` and `back` before changing it and re-assigning it after the call to `set_console`. What do you think?

andyneff:
@wiggin15 Yes to second one should be BRIGHT_BACKGROUND
@wiggin15 Are you suggesting adding another attribute for Windows to handle light? Upon reflection of how it does work in Linux, this makes a lot of sense (I think this is what tartley/colorama#20 was talking about towards the end)

```python
import colorama

colorama.init()

print colorama.Style.BRIGHT+'normal' + colorama.Fore.RED + 'red' \
      + colorama.Fore.RESET + 'normal' + colorama.Fore.LIGHTRED_EX \
      + 'light red' + colorama.Fore.RESET +'normal' + colorama.Style.NORMAL
print 'normal' + colorama.Fore.RED + 'red' + colorama.Fore.RESET + 'normal' \
      + colorama.Fore.LIGHTRED_EX + 'light red' + colorama.Fore.RESET +'normal'
print colorama.Style.BRIGHT+'normal' + colorama.Back.RED + 'red' \
      + colorama.Back.RESET + 'normal' + colorama.Back.LIGHTRED_EX \
      + 'light red' + colorama.Back.RESET +'normal' + colorama.Style.NORMAL
print 'normal' + colorama.Back.RED + 'red' + colorama.Back.RESET + 'normal' \
      + colorama.Back.LIGHTRED_EX + 'light red' + colorama.Back.RESET +'normal'

print colorama.Fore.RED + colorama.Style.BRIGHT + "this should be bright" \
      + colorama.Fore.RESET
print colorama.Fore.RED + colorama.Style.BRIGHT + "this should be bright" \
      + colorama.Fore.RESET
print colorama.Fore.RED + "this should still be bright" + colorama.Fore.RESET
print colorama.Fore.RED + "this should still be bright" + colorama.Fore.RESET

```
Now produces the desirable results?
windows2

andyneff/colorama@301523fe76201c1450f97ef6098b6f17d8b054fd

wiggin15:
I came up with something a little simpler: https://github.com/wiggin15/colorama/commit/28d6f63468e0dcfd6b431b3a78bf617f08d1dcd0 .
Can you please check the branch at: https://github.com/wiggin15/colorama
and let me know if that works correctly on all the cases above? Thanks.

andyneff:
Works perfectly!

Thanks

andyneff:
Well, that does work perfectly for the test lines I wrote... HOWEVER...

Adding this to the existing tests
```python
print colorama.Style.NORMAL

print 'normal' + colorama.Fore.LIGHTRED_EX + "light red" + colorama.Style.NORMAL\
      + "what am I" + colorama.Fore.RESET
```
In Linux tartley/colorama@5ab7cdd96b72d5bdb0fb25fdb5ac6ddc8cd7bb1f I get
linux2

But in Windows wiggin15/colorama@28d6f63468e0dcfd6b431b3a78bf617f08d1dcd0
windows3

And in Windows andyneff/colorama@301523fe76201c1450f97ef6098b6f17d8b054fd
windows4

Now, I would argue the behavior in Linux makes sense, because in my head, I'm treating "BRIGHT" and "LIGHT" as separable attributes (that just happen to be the same in Windows due to... well Windows, you know)

wiggin15:
The problem here is that the *LIGHT_EX colors change the style, but Back/Fore.RESET only reset the color and not the style.
We need Back/Fore.RESET to reset the style, but only if the style was changed due to the LIGHT_EX colors.

I have to agree that a separate member ('self._light' in this PR) is needed, because there is no
other way to remember if the style was changed due to Style.* or due to *LIGHT_EX.

andyneff:
I believe I agree. So will 301523fe76201c1450f97ef6098b6f17d8b054fd do the trick, or is something still missing?

wiggin15:
Can you just add a comment please explaining what self._light is? (that it is complementary to _style and is used for the LIGHT_EX colors, and we have to use it to separate the RESETs).

wiggin15:
Great. I'll merge this soon. Thanks.

PR #65, opened 2015-07-09
Updated README.rst

dannguyen:
removed redundant RESET calls in example, as Style.RESET_ALL is said to perform Fore.RESET and Back.RESET


wiggin15:
I think the idea was to show that you can reset each component separately, but the way it is now is probably more confusing than helping, so this is good. Thanks

Issue #51, opened 2015-04-22
support context manager syntax


tartley:
Reported by steve at the domain 3xile dot com, Aug 8, 2011
It'd be nice to turn colorized output on and off with `with` rather than either init() or writing by hand, especially when other modules want to modify sys.stdout.  Here's an example implementation:

    import sys
    from colorama import Fore, AnsiToWin32

    class ConvertedOutput(object):
        def __init__(self, *AnsiToWin32Args, **kwargs):
            self._AnsiArgs = AnsiToWin32Args
            self._set_stderr = kwargs.get('stderr', None)
        def __enter__(self):
            self._stdout = sys.stdout
            sys.stdout   = AnsiToWin32(sys.stdout, *self._AnsiArgs)
            if self._set_stderr:
                sys.stderr = AnsiToWin32(sys.stderr, *self._AnsiArgs)
        def __exit__(self, type, value, traceback):
            sys.stdout = self._stdout

And example usage:

    with ConvertedOutput():
        print('%sThis text is red%s' % (Fore.RED, Fore.RESET))

    print('%sThis text is surrounded by junk%s' % (Fore.RED, Fore.RESET))

Migrated from https://code.google.com/p/colorama/issues/detail?id=28&can=1

tartley:
Seems nice-to-have, but on the other hand, simple is good. Plus, I'm not sure of the use case when you want to have precise control over when conversion gets turned off, other than just program exit.

stevenhoward:
It's been a few years now, but I believe in this case I was working with another library that also wanted to modify `sys.stdout`, and I ran into some issues that I've long forgotten.

tartley:
I see. That makes sense then. Thanks Steven.

wiggin15:
Fixed by #67.

PR #67, opened 2015-08-26
Add context manager for colorama

digwanderlust:
This enables you to handle the stdout replacement with a with statement rather than manually keeping track of it... for example:

with colorama_text():
      print Fore.GREEN, 'text is green'
      print Fore.RESET, 'text is back to normal'

print 'text is back to stdout'

### DIFF
--- a/colorama/initialise.py
+++ b/colorama/initialise.py
@@ -56,8 +56,10 @@ def deinit():
 @contextlib.contextmanager
 def colorama_text(*args):
     init(*args)
-    yield
-    deinit()
+    try:
+        yield
+    finally:
+        deinit()
 
 def reinit():
     if wrapped_stdout is not None:
--- a/colorama/initialise.py
+++ b/colorama/initialise.py
@@ -18,6 +18,7 @@ atexit_done = False
 def reset_all():
     AnsiToWin32(orig_stdout).reset_all()
 
+
 def init(autoreset=False, convert=None, strip=None, wrap=True):
 
     if not wrap and any([autoreset, convert, strip]):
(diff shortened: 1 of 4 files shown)


digwanderlust:
Are there other steps I should follow to get this merged?

wiggin15:
This looks good to me (I wasn't sure about the name of the context method but I guess it's alright). This will also resolve #51.

colorama/initialise.py
```diff
@@ -17,7 +18,6 @@
 def reset_all():
     AnsiToWin32(orig_stdout).reset_all()
 
-
```
wiggin15: There are two blank lines between functions in this file, so this line shouldn't be removed

colorama/initialise.py
```diff
@@ -53,6 +53,14 @@ def deinit():
...
 
 
+@contextlib.contextmanager
+def colorama_text(*args):
+    init(*args)
+    try:
+        yield
+    finally:
+        deinit()
+
```
wiggin15: There should be two blank lines following this function too

colorama/initialise.py
```diff
@@ -53,6 +53,14 @@ def deinit():
         sys.stderr = orig_stderr
 
 
+@contextlib.contextmanager
+def colorama_text(*args):
```
wiggin15: Maybe we should add **kwargs too to pass to 'init' so we can use the context manager with the keywords

wiggin15:
Sorry for the new batch of comments. I should have done a better review before. Thanks for fixing.

digwanderlust:
No worries, just finished up the changes... let me know if you see anything
else.

wiggin15:
Thanks!

PR #68, opened 2015-09-11
Fix false-positive for windows detection

mhils:
I'm using bash on Windows as my main console (via Cygwin/[Git Bash](https://git-for-windows.github.io/)). However, colorama treats every system that has `$TERM` defined as an emulated Windows - Cygwin environments should be excluded from that.

Thanks for the great work you're doing here! :smiley:

### DIFF
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -56,7 +56,7 @@ class AnsiToWin32(object):
         self.stream = StreamWrapper(wrapped, self)
 
         on_windows = os.name == 'nt'
-        on_emulated_windows = on_windows and 'TERM' in os.environ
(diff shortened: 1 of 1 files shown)


wiggin15:
Actually, the `$TERM` check was added specifically for Cygwin. The Cygwin terminal doesn't support the Windows API that colorama uses and does support parsing the ANSI escape codes, so on Cygwin we want to use the ANSI escape codes (i.e., convert=False).
Git Bash uses Cygwin's bash on cmd (not on the Cygwin terminal), and should support both conversion methods. For some reason it doesn't convert the escape codes when they are printed by colorama (it does convert them when using, e.g., 'echo').
Please see issue #48. There should be a different fix for the Git Bash problem.

mhils:
Sorry, I did not notice #48 - closed then.

PR #69, opened 2015-09-18
python3 fix for demo07.py

alekibango:


wiggin15:
Thanks for the patch. The file is missing "from __future__ import print_function" so this wouldn't print correctly on Python 2.
I actually committed a fix for both this demo and demo08 a few days ago, and just didn't push it. It's pushed now.

wiggin15:
(I was mistaken - it does print correctly on Python 2, but it's still better to have the print_function import)

Issue #71, opened 2015-09-25
Finalise formal project ownership handover


tartley:
@wiggin: My fingerprints are still all over this project, but I never use Windows these days, plus am now a Dad, so these days I have little time, and even less incentive, to contribute.

This ticket is my attempt to clean up any vestiges of my ownership that might make life less pleasant for @wiggin and other future contributors. Things that *might* need changing (let's discuss here):

* Does it make sense for the github repo to continue to be owned by me (tartley)? Or should we change it to @wiggin15? Or should we create a new 'colorama' user or organisation to encourage contributions & group ownership? The downside of changing is that (I assume) it changes the repo URL (again.) I think it's fine as-is, but would love to hear people's thoughts.
* Does it make sense to transfer ownership of the PyPI entry? Remind me, did I give anyone the credentials needed to make a release without my input? If not, let's fix that somehow.
* The README lists me as the only entry under 'Contacts'. This no longer makes sense. I do occasionally get bug reports or requests for help to that address (which I try to forward on to the proper places.) Should we replace it with contact details for @wiggin15? Or just the Colorama mailing list? (PR https://github.com/tartley/colorama/pull/72 uses the mailing list)
* Anything else?

wiggin15:
Hi. First of all, I still think of you as the owner of this project (I only help maintain it ;))
It's good that you're still involved, even if it's only by getting notifications about issue updates.
* Regarding PyPI, you did give me permissions to upload new releases. I wanted to ask you if it's ok if I release new versions when needed.
* Regarding contact information: I think that the README shouldn't list email addresses, so that discussion and bug reports will be more public. The mailing list has little to no activity, and I'm not sure it's needed. I think here (GitHub) is the best place for discussion and reporting issues, so maybe we only need a link to this repository.
* Regarding moving the repository: it's a good idea, but it's probably more trouble than it is worth. Keeping it as-is is fine as far as I'm concerned.

kirbyfan64:
Or move the repo to a colorama organization. I'm pretty sure GitHub keeps the original repo links intact.

tartley:
acknowledged, makes sense.

@wiggin Yep, absolutely, release as required.

A modified PR for the README with contact links on its way then...

Hugs!

nmstoker:
[...]

PR #72, opened 2015-09-25
Issue71 readme tidy

tartley:
A start towards addressing https://github.com/tartley/colorama/issues/71

Most of the diffs are contentless whitespace faffing - I like to view things in an 80 char terminal. My apologies.

The only substantive changes are:
* I removed my own email address as a suggested project 'contact' (because I no longer actively contribute these days)
* I removed contributor Jesse's email address from this public README. (Sorry Jesse!)
* I added the word 'contact' to the link to the colorama discussion group, so that people doing 'find in page' for 'contact' will find it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -69,7 +69,8 @@ Copyright Jonathan Hartley 2013. BSD 3-Clause license; see LICENSE file.
 Dependencies
 ============
 
-None, other than Python. Tested on Python 2.5.5, 2.6.5, 2.7, 3.1.2, 3.2, 3.3, and 3.4.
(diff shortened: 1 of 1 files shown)


wiggin15:
Thanks! More things to consider:
* Remove this line: "Ask Jonathan for commit rights, you'll get them."
* Remove the Thanks section? Of course I don't want to remove any ACK to contributors, but most of this list duplicates entries in the changelog, and I'm not sure the README is the place for this. Maybe we need to find the names not mentioned in the changelog, find how they contributed, and move them to the changelog. Alternatively, create a separate file for ACKs (I saw other projects using it, e.g. psutil and Python itself). This is just a suggestion - what do you think?
* Your name should be more prominent than a single line at the bottom :)

PR #74, opened 2015-10-09
Add check if orig_stdout is not None before calling AnsiToWin32::reset_all()

PrzemekWirkus:
# Description
When calling ```AnsiToWin32(orig_stdout).reset_all()``` inside ```reset_all()``` function (https://github.com/tartley/colorama/blob/master/colorama/initialise.py#L18) we assume that ```orig_stdout``` is not None, which is not true for all cases.

This causes issues, example below:
```
$ cd python-package-sources
$ coverage -x setup.py test
...
----------------------------------------------------------------------
Ran 15 tests in 0.013s

OK
Error in atexit._run_exitfuncs:
Traceback (most recent call last):
  File "C:\Python27\lib\atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "C:\Python27\lib\site-packages\colorama\initialise.py", line 18, in reset_all
    AnsiToWin32(orig_stdout).reset_all()
TypeError: 'NoneType' object is not callable
Error in sys.exitfunc:
Traceback (most recent call last):
  File "C:\Python27\lib\atexit.py", line 24, in _run_exitfuncs
    func(*targs, **kargs)
  File "C:\Python27\lib\site-packages\colorama\initialise.py", line 18, in reset_all
    AnsiToWin32(orig_stdout).reset_all()
TypeError: 'NoneType' object is not callable
```

### DIFF
--- a/colorama/initialise.py
+++ b/colorama/initialise.py
@@ -16,7 +16,8 @@ atexit_done = False
 
 
 def reset_all():
-    AnsiToWin32(orig_stdout).reset_all()
+    if orig_stdout:
+        AnsiToWin32(orig_stdout).reset_all()
 
 
 def init(autoreset=False, convert=None, strip=None, wrap=True):


wiggin15:
Thanks for reporting this issue.
I'm not sure in what case orig_stdout is None.
reset_all is registered to be called at exit only after orig_stdout is assigned. Nothing assigns it back to None.
In the traceback you pasted it is also not clear that orig_stdout is the problem (we're not trying to call it).
Where does the script 'coverage' come from? I'm trying to reproduce.

PrzemekWirkus:
Hi,
Regarding coverage (Python code coverage tool):
```
$ coverage --version
Coverage.py, version 3.7.1.  http://nedbatchelder.com/code/coverage
```

PrzemekWirkus:
Hi,
I've updated Python and few packages on my system. I no longer can observe this issue on my system.
Should we close this PR?

wiggin15:
I am able to reproduce this. Looks like `AnsiToWin32` becomes `None` when atexit is called.

wiggin15:
@PrzemekWirkus can you please change the `if` to check `AnsiToWin32` instead of `orig_stdout`, and also add a comment that explains that imported objects may be become None when the program closes? Thanks.

Here's a related question from StackOverflow: http://stackoverflow.com/questions/2572172/referencing-other-modules-in-atexit

PR #100, opened 2016-06-21
Add light colors in README

Timokasse:
Let's not hide those in the dark of the code.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -143,6 +143,11 @@ Available formatting constants are::
 ``Style.RESET_ALL`` resets foreground, background, and brightness. Colorama will
 perform this reset automatically on program exit.
 
+These are fairly well supported, but not part of the standard::
+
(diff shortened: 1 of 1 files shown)


wiggin15:
Hiding these colors from the documentation was intentional. Since they are not part of the standard, it is not recommended to use them, and using AnsiStyle.BRIGHT is a better idea. The EX colors are there more because we want colorama to know how to convert them to Windows API calls _if_ they are used, but they should not really be used in the code. See https://github.com/tartley/colorama/pull/20.

I'm open to other opinions, though, if you still think it should be documented.

Timokasse:
If there is a better way, I am all for it. But I tried and I don't see how I could do a
`print(Fore.LIGHTRED_EX + 'some light red text')`
with standard colors.

Also, I have specified that these colors are not part of the standard because I believe it is best to provide accurate information. And since it works on Linux and Cygwin (Windows) at least, I am convinced it could be useful to many.

tartley:

PR #105, opened 2016-08-15
Add Win 10 native ANSI cmd support

Slayther:


Slayther:
Not sure why `on_windows` check didn't prevent `check_win_10_ansi_support` from running..

colorama/ansitowin32.py
```diff
@@ -21,6 +21,20 @@ def is_a_tty(stream):
     return hasattr(stream, 'isatty') and stream.isatty()
 
 
+def check_win_10_ansi_support():
```
wiggin15: Is it possible to replace the windows release check using the "platform" module? platform.win32_ver() or platform.platform()?

colorama/ansitowin32.py
```diff
@@ -66,13 +80,16 @@ def __init__(self, wrapped, convert=None, strip=None, autoreset=False):
         # to support the ANSI codes.
         conversion_supported = on_windows and winapi_test()
 
+        # Does this version of Win 10 support ANSI?
+        win_10_ansi_support = on_windows and check_win_10_ansi_support()
+
         # should we strip ANSI sequences from our output?
-        if strip is None:
+        if strip is None and not win_10_ansi_support:
```
wiggin15: I prefer to add the win10 check inside the if, this way "strip" will always be True/False and not be left as None. Same for 'convert'

colorama/ansitowin32.py
```diff
@@ -66,13 +80,16 @@ def __init__(self, wrapped, convert=None, strip=None, autoreset=False):
         # to support the ANSI codes.
         conversion_supported = on_windows and winapi_test()
 
+        # Does this version of Win 10 support ANSI?
+        win_10_ansi_support = on_windows and check_win_10_ansi_support()
```
wiggin15: Let's change the name to "windows_ansi_support", in case it will be more than just Windows 10 in the future :)

colorama/ansitowin32.py
Slayther: I haven't found a way to check the build version using the platform module.

colorama/ansitowin32.py
Slayther: Ok

Slayther:
Ok, I did the changes you suggested. Check it out and see if that is what you wanted.

colorama/ansitowin32.py
kirbyfan64: @Slayther Try `sys.getwindowsversion()[2]`.

colorama/ansitowin32.py
```diff
@@ -69,14 +86,17 @@ def __init__(self, wrapped, convert=None, strip=None, autoreset=False):
...
         # should we strip ANSI sequences from our output?
         if strip is None:
-            strip = conversion_supported or (not is_stream_closed(wrapped) and not is_a_tty(wrapped))
+            strip = conversion_supported or not windows_ansi_support or (not is_stream_closed(wrapped) and not is_a_tty(wrapped))
         self.strip = strip
 
         # should we should convert ANSI sequences into win32 calls?
         if convert is None:
-            convert = conversion_supported and not is_stream_closed(wrapped) and is_a_tty(wrapped)
+            convert = not windows_ansi_support or (conversion_supported and not is_stream_closed(wrapped) and is_a_tty(wrapped))
```
wiggin15: Are you sure this should be "or"-ed and not "and"-ed? This way if the old condition with conversion_supported is True, then convert will be True even if windows_ansi_support is True, which (I think) is not the desired outcome. I think that there is the same problem with strip (but I may be mistaken). This is a difficult part.

colorama/ansitowin32.py
```diff
@@ -21,6 +21,23 @@ def is_a_tty(stream):
     return hasattr(stream, 'isatty') and stream.isatty()
 
 
+def check_windows_ansi_support():
+    try:
+        import winreg
```
wiggin15: There is a problem here because winreg only exists in Python 3 but we support Python 2 too. Also, this function fails on older versions of Windows because the ReleaseID value doesn't exist in the registry. I am reluctant to pull this function - too much platform specific code has a high chance of breaking things.

kirbyfan64:
I mentioned this in the review comments, but, just in case it got skipped over:

`sys.getwindowsversion()[2]` can be used to retrieve the Windows build number on Python versions as far back as 2.3. So `check_windows_ansi_support` can just be:

```python
def check_windows_ansi_support():
    return sys.getwindowsversion()[2] >= 1511
```

wiggin15:
On Windows 7 this returns True. The check should include the major and minor, not just the build. A tuple comparison should work.

colorama/ansitowin32.py
Slayther: I worked with Python 2 when I encountered the error and fixed it with this check function. Though, I agree it contains a lot of platform specific code, but I don't know what else would work.

colorama/ansitowin32.py
Slayther: I'm not really sure. That's why I put the check on the if line at first. I tried several combinations, but couldn't find one that fits for all cases.

colorama/ansitowin32.py
Slayther: I'm not sure why I can't import it anymore. I am sure it worked in Python 2.7.x

colorama/ansitowin32.py
Slayther: I think we should have separate ifs, one for on_windows and one for other platforms.

colorama/ansitowin32.py
MinchinWeb: The module is named `_winreg` on Python 2.7 (notice the leading underscore). https://docs.python.org/2/library/_winreg.html

MinchinWeb:
This would be wonderful to have! Powershell (on Windows 10) [now supports 24 bit color](https://blogs.msdn.microsoft.com/commandline/2016/09/22/24-bit-color-in-the-windows-console/), but when colorama is initialized, we get dropped back to 16 color mode.

Below is the same script, the first run without colorama, the second with colorama initialized.

image

MinchinWeb:
Adding to the complexity, this new functionality only works on Powershell, and not on CMD...

segevfiner:
Are you sure this only works in powershell? I think you simply need to enable `ENABLE_VIRTUAL_TERMINAL_PROCESSING`.

image


0x4 is `ENABLE_VIRTUAL_TERMINAL_PROCESSING`. cmd has a weird behavior that powershell doesn't have. It enables `ENABLE_VIRTUAL_TERMINAL_PROCESSING` on start, disables it when it starts a child program and doesn't disable it on exit (If you start cmd as a child of Python it will leak `ENABLE_VIRTUAL_TERMINAL_PROCESSING`).

mixmastamyk:
@kirbyfan64 

The build number is not the same as the ReleaseID, but they correspond:

- https://en.wikipedia.org/wiki/Windows_10_version_history

So the code should look like this:

def check_windows_ansi_support():
    return sys.getwindowsversion()[2] >= 10586

tartley:

tartley:
Also, we'd need corresponding test changes before this could merge.

njsmith:
#139 uses a more direct/reliable strategy to handle the ANSI stuff, so closing this in favor of it

Issue #107, opened 2016-10-07
Can not display color in PyCharm using default init()


aploium:
[PyCharm](https://www.jetbrains.com/pycharm/) is a widely used python IDE, which has a built-in python console.
However, the default init() function doesn't works very well in it's console. See example below:

``` python
import colorama

colorama.init()
print(colorama.Fore.GREEN + "hello world-green" + colorama.Fore.RESET)
print(colorama.Fore.RED + "hello world-red" + colorama.Fore.RESET)
```

![sc 2016-10-07_16-53-03](https://cloud.githubusercontent.com/assets/15957250/19184567/678437d0-8cb0-11e6-8401-6bebf2144ace.png)

And we have to set `convert=False` and `strip=False` to make it works in PyCharm console

``` python
import os
import colorama

if 'PYCHARM_HOSTED' in os.environ:
    convert = False  # in PyCharm, we should disable convert
    strip = False
    print("Hi! You are using PyCharm")
else:
    convert = None
    strip = None

colorama.init(convert=convert, strip=strip)
print(colorama.Fore.GREEN + "hello world-green" + colorama.Fore.RESET)
print(colorama.Fore.RED + "hello world-red" + colorama.Fore.RESET)
```

![sc 2016-10-07_16-54-21](https://cloud.githubusercontent.com/assets/15957250/19184604/90123e18-8cb0-11e6-9db3-c8e82b3fe9fa.png)

> **PYCHARM_HOSTED**  
> It's an special environment that only occurs in PyCharm,
> see http://stackoverflow.com/questions/29777737/how-to-check-if-python-unit-test-started-in-pycharm-or-not

proximous:
Has this improvement made it into any official release? If not, is there a plan to include it or a reason not to? This is a nice feature I'd like to use. Thanks.

Soaku:
Doesn't work for me 😕

PR #108, opened 2016-10-07
Add PyCharm support

aploium:
Please see Issue #107


wiggin15:
Hi. I think the check will fit more neatly in ansitowin32.py where the default strip/convert values are determined. Maybe we can add the PyCharm check to `conversion_supported`, or directly to the lines that determine strip and convert: https://github.com/tartley/colorama/blob/master/colorama/ansitowin32.py#L74 and https://github.com/tartley/colorama/blob/master/colorama/ansitowin32.py#L79

PR #109, opened 2016-10-11
avoid ansi-escape replacing inside eclipse-pydev

x2nie:
avoid conflicting colorama with the recommended eclipse plugin (ansi-escape) by pydev author that can enable color in pydev-console
it's make both running code in terminal & eclipse console can show the correct color
image

### DIFF
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -8,6 +8,7 @@ from .winterm import WinTerm, WinColor, WinStyle
 from .win32 import windll, winapi_test
 
 
+INSIDE_PYDEV = 'sitecustomize' in sys.modules #for avoid ansi-escape replacing when runngin inside eclipse-pydev
 winterm = None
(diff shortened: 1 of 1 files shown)


wiggin15:
Hi. I think the check will fit more neatly in `AnsiToWin32.__init__` where the default strip/convert values are determined (I think we need convert=False and strip=False). Maybe we can add the pydev check to `conversion_supported`, or directly to the lines that determine strip and convert: https://github.com/tartley/colorama/blob/master/colorama/ansitowin32.py#L74 and https://github.com/tartley/colorama/blob/master/colorama/ansitowin32.py#L79

In addition there are pep8 issues with the comments:
```
colorama/ansitowin32.py:11:46: E261 at least two spaces before inline comment
colorama/ansitowin32.py:11:47: E262 inline comment should start with '# '
```

tartley:

PR #113, opened 2016-11-14
Adding new print helper function (Word) in ansi module

mkusz:
Adding new function that can help formatting strings when you try to use print with .format (sometimes it's much simpler than performing multiple printing or writing own functions for this. Example is inside new README.rst file. There are also some additional tests for this functionality.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -120,6 +120,14 @@ constant shorthand for ANSI escape sequences:
     print('\033[31m' + 'some red text')
     print('\033[30m') # and reset to default color
 
+...or by using ``Word()`` when you want to use ``.format()`` like this:
+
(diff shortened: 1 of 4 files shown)


wiggin15:
Hi. Thanks for the pull request! I think the name of the function you're proposing is a little confusing, because
* "Word" can accept any string, even if it's more than one word
* The capital W makes the function look like a class because of the default Python conventions (PEP 8)
* It's not immediately clear what is the purpose of the function from its name
Maybe changing the name to `add_color`/`colorize` or something along these lines is a good idea? It looks very much like the `colored` function from the `termcolor` package.

tartley:

PR #129, opened 2017-04-21
Support foreground/background reversing

zwhfly:
ESC [ 7 m    swap foreground and background


wiggin15:
Thanks for this pull request. This looks good.
One thing I noticed is that on my Mac terminal, writing the swap ansi code (ESC 7m) twice doesn't re-reverse the foreground and background but instead keeps them reversed. With your change, on Windows, writing the code twice reverts to the original state.
In addition, can you please add a helper function to print this code, like we have for example `colorama.ansi.clear_line`? e.g. add `colorama.ansi.inverse`. Thanks.

tartley:
Does this duplicate the functionality offered by #267?

tartley:
These code changes can't be merged until they also have corresponding test changes.

Issue #114, opened 2016-11-28
ANSI codes instead of colors when pytest imports colorama through conftest.py


grovduck:
I had reported this as an issue at [pytest](https://github.com/pytest-dev/pytest/issues/1515), but it's probably more likely an issue with `colorama`.  The gist of the issue is that when a user has a custom pytest `conftest.py` file used which, in turn, imports `colorama`, the `handles` dictionary within `win32.py` does not get reparsed on the second import.  I've suggested a possible fix in that ticket, but will defer to your judgment on how best to fix it.  Thanks for looking at it.

segevfiner:
This is caused by importing colorama with pytest's fd redirection active. I implemented a workaround for this here https://github.com/pytest-dev/pytest/pull/2611.

Issue #104, opened 2016-08-14
Support for Windows 10 built-in ANSI cmd


Slayther:
Windows 10 has built-in support for ANSI cmd. The ANSI codes should not be stripped.
The module should check for this. I know how to, but I don't know where to put this check..

wiggin15:
Helo @Slayther. Is colorama broken on Windows 10 cmd, or do both coloring methods (ANSI codes and API calls) work?

The decision about whether to strip and/or convert ANSI codes is in ansitowin32.py (currently lines 74 and 79). It would be great if you could submit a pull request. Otherwise please specify how to check for Windows 10 ANSI code support here. Thanks!

Slayther:
Well, both ANSI codes and API calls work on new Windows 10 cmd. I thought I saw the dim option working on the new cmd, but I was mistaken once I commented out both ANSI stripping and API calls. Everything else works as expected on Windows.

Nevertheless, if Windows updates support for ANSI sequences, it would be good to support those right away.

I will submit the pull request shortly, once I test some stuff out.

Slayther:
Fixed with #105.

vidartf:
While simple test runs of colorama works well on the new Windows 10 cmd, more complex usage of colorama breaks. E.g. calling py.test works well, calling py.test with a specific file gives sometimes black text on black background (also, Ctrl+C leaves console in whatever color state it was in when exception occurred). Other applications which use colorama also act strange. I unfortunately don't know enough about how colorama works to create a minimal working example.

But yeah, +1 from me on getting a fix in :)

vidartf:
As colorama is broken in many cases for me on Windows 10, and the submitted PR is still open, I would propose to reopen this issue.

Slayther:
I did not check the PR as this was a temporary fix before I started using the key codes themselves. @wiggin15 As the build is passing and the features is proven to work, could you pull the request?

wiggin15:
@vidartf it would be very helpful if you can produce minimal test cases anyway so we can find exactly what is broken. The behavior on Windows 10 should not be different than the behavior on other Windows versions if using the API for color conversion. I think if we can leave things the way they are without adding detection for Windows 10 it would be better.

vidartf:
Lifting the relevant bits from `py.test`, I was able to come up with a minimal example:
```python
import sys
import os
from tempfile import TemporaryFile


# Capture stdout
stdoutfd = 1
saved_fd = os.dup(stdoutfd)
[...]

PR #131, opened 2017-06-02
Don't break due to our handles getting closed by FD redirection

segevfiner:
The fix is to simply not cache the std handles returned by `GetStdHandle`

This happens in Pytest on at least Python 3.5 and 3.6.
Please check https://github.com/pytest-dev/pytest/issues/2465 for more information about this before merging.

### DIFF
--- a/colorama/win32.py
+++ b/colorama/win32.py
@@ -89,11 +89,6 @@ else:
     ]
     _SetConsoleTitleW.restype = wintypes.BOOL
 
-    handles = {
-        STDOUT: _GetStdHandle(STDOUT),
-        STDERR: _GetStdHandle(STDERR),
-    }
-
     def _winapi_test(handle):
         csbi = CONSOLE_SCREEN_BUFFER_INFO()
         success = _GetConsoleScreenBufferInfo(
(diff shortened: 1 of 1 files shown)


wiggin15:
This could be related to #114 and maybe https://github.com/tartley/colorama/issues/104#issuecomment-261772454 is also slightly related

segevfiner:
Arghhhh, I started seeing this on Python 2.7 too. Who knows what changed that it's now triggered there too. Be it Pytest version, Windows version, or what not... This PR seems to fix it for that instance too, anyhow.

Victorious3:
I would love to see this one fixed as well, ran into it when using pytest.

nicoddemus:
> I would love to see this one fixed as well, ran into it when using pytest.

I second that. @wiggin15 anything missing here? Anything else we can help with to get this merged and a new PyPI release?

nicoddemus:
Great, thanks a lot @wiggin15!

wiggin15:
Just to make sure - is the problem in pytest related to stream _redirection_?
Any chance you can summarize the problem in one sentence for the changelog - something like "don't cache the std handles to avoid problems where/with ..."
Thanks.

segevfiner:
> Just to make sure - is the problem in pytest related to stream _redirection_?
> Any chance you can summarize the problem in one sentence for the changelog - something like "don't cache the std handles to avoid problems where/with ..."
> Thanks.

Yes. It is caused by FD redirection (`dup2`). Windows is closing the underlying handle and they don't always end up with the same number when reopened. Maybe something like: "Don't cache stdio HANDLEs, since they might be closed/changed by fd redirection".

wiggin15:
Thanks. There's a new release, 0.4.0, with this change.

nicoddemus:
Awesome, thanks @wiggin15 for the prompt merge and release! 🙇

PR #133, opened 2017-06-11
Provide pass-through ANSI on Windows 10 powershell

MinchinWeb:
This provides pass-thru to ANSI codes on new enough Windows 10 versions, and to Powershell only. It does this by using `psutil` to check the process name that is running python, and if that matches `powershell.exe`, ANSI pass-thru is enabled.

This does add `psutil` as a dependency on Windows.

Closes #105 

Basic local testing on both Python 2.7 and Python 3.6

### DIFF
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -21,7 +21,7 @@ def is_a_tty(stream):
     return hasattr(stream, 'isatty') and stream.isatty()
 
 
-def check_win_10_ansi_support():
+def check_windows_ansi_support():
     try:
         import winreg
     except ImportError:
@@ -87,16 +87,16 @@ class AnsiToWin32(object):
         conversion_supported = on_windows and winapi_test()
 
         # Does this version of Win 10 support ANSI?
-        win_10_ansi_support = on_windows and check_win_10_ansi_support()
+        windows_ansi_support = on_windows and check_windows_ansi_support()
 
         # should we strip ANSI sequences from our output?
-        if strip is None and not win_10_ansi_support:
-            strip = conversion_supported or (not is_stream_closed(wrapped) and not is_a_tty(wrapped))
+        if strip is None:
+            strip = conversion_supported or not windows_ansi_support or (not is_stream_closed(wrapped) and not is_a_tty(wrapped))
         self.strip = strip
 
         # should we should convert ANSI sequences into win32 calls?
-        if convert is None and not win_10_ansi_support:
-            convert = conversion_supported and not is_stream_closed(wrapped) and is_a_tty(wrapped)
+        if convert is None:
+            convert = not windows_ansi_support or (conversion_supported and not is_stream_closed(wrapped) and is_a_tty(wrapped))
         self.convert = convert
 
         # dict of ansi codes to win32 functions and parameters
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -22,7 +22,10 @@ def is_a_tty(stream):
 
 
 def check_win_10_ansi_support():
-    import winreg
+    try:
+        import winreg
+    except ImportError:
+        return False
 
     # Check the registry for release ID
     key = r"SOFTWARE\Microsoft\Windows NT\CurrentVersion"
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -25,7 +25,12 @@ def check_windows_ansi_support():
     try:
         import winreg
     except ImportError:
-        return False
+        try:
+            import _winreg  # on Python 2
+        except ImportError:
+            return False
+        else:
+            winreg = _winreg
 
     # Check the registry for release ID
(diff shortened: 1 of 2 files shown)


MinchinWeb:
Pictures!

CMD (so only 16 color mode, but that works):

image

Powershell (full 256 color glory!):

image

segevfiner:
@MinchinWeb Are you sure this only works in powershell? I think you simply need to enable `ENABLE_VIRTUAL_TERMINAL_PROCESSING`.

image


0x4 is `ENABLE_VIRTUAL_TERMINAL_PROCESSING`. cmd has a weird behavior that powershell doesn't have. It enables `ENABLE_VIRTUAL_TERMINAL_PROCESSING` on start, disables it when it starts a child program and doesn't disable it on exit (If you start cmd as a child of Python it will leak `ENABLE_VIRTUAL_TERMINAL_PROCESSING`).

Reference: https://msdn.microsoft.com/en-us/library/windows/desktop/ms686033(v=vs.85).aspx

colorama/ansitowin32.py
```diff
@@ -21,6 +21,35 @@ def is_a_tty(stream):
...
+
+    # Check the registry for release ID
+    key = r"SOFTWARE\Microsoft\Windows NT\CurrentVersion"
+    val = r"ReleaseID"
+    key = winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE, key)
+    releaseId = int(winreg.QueryValueEx(key,val)[0])
+    winreg.CloseKey(key)
+
+    # Windows 10 supports ANSI cmd since release 1511
+    if releaseId >= 1511:
```
segevfiner: You probably need to check that this is Windows 10 too, I'm not sure ReleaseID isn't used in previous versions of Windows. l Windows since 8.1 has a nasty habit of lying about it's version though. I think `platform.win32_ver()` does work around that but be sure to check (`sys.getwindowsversion` doesn't workaround this at least on Python 2.7.13 so you get the version of Windows 8).

colorama/ansitowin32.py
```diff
@@ -21,6 +21,35 @@ def is_a_tty(stream):
...
+        # but only on powershell
+        import psutil
```
segevfiner: I think this isn't really related to powershell but to the `ENABLE_VIRTUAL_TERMINAL_PROCESSING` flag. colorama should enable this flag when supported and restore it's previous value on exit.

I think we can simply try to `SetConsoleMode(h, GetConsoleMode(h) | ENABLE_VIRTUAL_TERMINAL_PROCESSING)` to check for support of this (Need to be sure the stream is a tty/console before calling `{Get,Set}ConsoleMode`). I think it fails with `INVALID_ARGUMENT` when unsupported. But this needs some testing.

See: https://github.com/tartley/colorama/pull/133#issuecomment-315325623 & https://msdn.microsoft.com/en-us/library/windows/desktop/ms686033(v=vs.85).aspx

setup.py
```diff
@@ -64,6 +64,9 @@ def get_version(path):
         'Programming Language :: Python :: 3.4',
         'Programming Language :: Python :: 3.5',
         'Topic :: Terminals',
-    ]
+    ],
+extras_require={
```
segevfiner: Indentation is messed up.

colorama/ansitowin32.py
```diff
@@ -69,14 +98,23 @@ def __init__(self, wrapped, convert=None, strip=None, autoreset=False):
...
         conversion_supported = on_windows and winapi_test()
 
+        # Does this version of Win 10 support ANSI?
+        windows_ansi_support = on_windows and check_windows_ansi_support()
+
         # should we strip ANSI sequences from our output?
         if strip is None:
-            strip = conversion_supported or (not is_stream_closed(wrapped) and not is_a_tty(wrapped))
+            if windows_ansi_support:
+                strip = False
```
segevfiner: You still need to check that the stream isn't closed and is a tty otherwise we will skip stripping ansi codes with redirected output. Be careful, all those boolean conditions are quite confusing.

colorama/ansitowin32.py
```diff
@@ -21,6 +21,35 @@ def is_a_tty(stream):
...
+            import _winreg  # on Python 2
+        except ImportError:
+            return False
+        else:
+            winreg = _winreg
+
+    # Check the registry for release ID
+    key = r"SOFTWARE\Microsoft\Windows NT\CurrentVersion"
+    val = r"ReleaseID"
+    key = winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE, key)
```
segevfiner: Use `try finally` (It is a context manager since 2.6 but I think colorama still declares compatibility with Python 2.5).

MinchinWeb:
I think that if #139 is robust, that would be preferred just because it doesn't add another dependency to the project. But I'd be happy to see either merged!

njsmith:
#139 uses a more direct/reliable strategy to handle the ANSI stuff, so closing this in favor of it

PR #137, opened 2017-07-13
Save ('s') and restore ('u') support added

kgeorgiy:
Includes test and demo plus minor refactoring of ansitowin32

### DIFF
--- a/colorama/winterm.py
+++ b/colorama/winterm.py
@@ -76,10 +76,12 @@ class WinTerm(object):
     def set_console(self, attrs=None, on_stderr=False):
         if attrs is None:
             attrs = self.get_attrs()
-        handle = win32.STDOUT
-        if on_stderr:
-            handle = win32.STDERR
+        handle = self.get_handle(on_stderr)
         win32.SetConsoleTextAttribute(handle, attrs)
+
+    @staticmethod
+    def get_handle(on_stderr=False):
+        return win32.STDERR if on_stderr else win32.STDOUT
 
     def get_position(self, handle):
         position = win32.GetConsoleScreenBufferInfo(handle).dwCursorPosition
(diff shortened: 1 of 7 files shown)


colorama/winterm.py
```diff
@@ -89,31 +91,38 @@ def get_position(self, handle):
...
+        handle = self.get_handle(on_stderr)
         win32.SetConsoleCursorPosition(handle, position)
 
     def cursor_adjust(self, x, y, on_stderr=False):
-        handle = win32.STDOUT
-        if on_stderr:
-            handle = win32.STDERR
-        position = self.get_position(handle)
-        adjusted_position = (position.Y + y, position.X + x)
+        (cy, cx) = self.get_cursor_position(on_stderr, adjust=False)
```
wiggin15: If we decide to continue to use get_position instead of the new get_cursor_position, we will not need this change.

tartley:

colorama/winterm.py
```diff
@@ -76,11 +76,13 @@ def style(self, style=None, on_stderr=False):
...
             attrs = self.get_attrs()
-        handle = win32.STDOUT
-        if on_stderr:
-            handle = win32.STDERR
+        handle = self.get_handle(on_stderr)
         win32.SetConsoleTextAttribute(handle, attrs)
 
+    @staticmethod
+    def get_handle(on_stderr=False):
+        return win32.STDERR if on_stderr else win32.STDOUT
```
tartley, line 84: I like this ^

crackwitz:
I could have used this feature right now. Anyone with privileges, please take another look.

JanSichula:
Please merge it. Reliable way to get cursor position will be very useful.

tartley:
Eight years late, but moving fast. Since this is on a fork's master, I'm attempting to fix merge conflicts, and address review comments, in #402

tartley:
Although, I have to ask: If one is not on Windows, then there's no reason to use Colorama, and my understanding is that modern Windows handles ANSI codes properly. So do people really still need new features adding to Colorama?

PR #141, opened 2017-08-24
Added colorama() context manager

geelenb:


geelenb:
So this pull enables functionality as seen in the (new) readme:

```
with colorama(Fore.RED):
    print('some red text')
    with colorama(Back.GREEN):
        print('and with a green background')
        with colorama(Style.DIM):
            print('and in dim text'):
print('back to normal now')
```

colorama/ansi.py
```diff
@@ -100,3 +101,33 @@ class AnsiStyle(AnsiCodes):
...
+    for arg in args:
+        print(arg, end='');
+        if arg in fore_values:
+            fore_stack.append(arg)
+        elif arg in back_values:
+            back_stack.append(arg)
+        elif arg in style_values:
+            style_stack.append(arg)
+    yield
+    for arg in reversed(args):
```
wiggin15: I think we don't need this loop. If we pass more than one `Fore` value, for example, we are going to print the colors in reverse (except the first one), but we don't need to print any color codes after the text has been written. The reason this code works is because the stacks always end with a reset - so all we need to do after the `yield` is print the reset codes, or only print `Style.RESET_ALL`.

colorama/ansi.py
```diff
@@ -100,3 +101,33 @@ class AnsiStyle(AnsiCodes):
...
+fore_values = set(Fore.__dict__.values())
+back_values = set(Back.__dict__.values())
+style_values = set(AnsiStyle.__dict__.values())
+
+fore_stack = [Fore.RESET]
+back_stack = [Back.RESET]
+style_stack = [Style.NORMAL]
+
+@contextlib.contextmanager
+def colorama(*args):
```
wiggin15: I think a better name would be `colorama_color` or something else. The name `colorama` is the name of the package and I don't think it's a good idea to override it. Also, we already have a `colorama_text` context manager which is something different, so it's better to avoid confusion.

geelenb:
Thanks for the feedback. I do believe Colorama could be greatly enhanced by context manager syntax, as the current syntax using string concatenation feels very unnatural. 

I also implemented another solution to this end, which I feel has the ideal syntactic solution, though its implementation is slightly worse. It would be used like this:

from colorama.contexts import Fore, Back, Style

```python
print('Before any, ', end='')
with Fore.RED:
    print('inside one, ', end='')
    with Back.BLUE:
        print('inside two, ', end='')
        with Fore.WHITE, Style.DIM:
            print('inside three, ', end='')
        print('nested ', end='')
    print('context ', end='')
print('managers.', end='')
```

And in `contexts.py`: 


```python
from colorama import ansi

fore_stack = [ansi.Fore.RESET]
back_stack = [ansi.Back.RESET]
style_stack = [ansi.Style.RESET_ALL]

class Context(object):
    def __init__(self, style, stack):
        self.style = style
        self.stack = stack

    def __enter__(self):
        print(self.style, end='')
        self.stack.append(self.style)

    def __exit__(self, exc_type, exc_val, exc_tb):
        if self.style == self.stack[-1]:
            self.stack.pop()
            print(self.stack[-1], end='')

class ForeContexts(object):
    BLACK = Context(ansi.Fore.BLACK, fore_stack)
    RED = Context(ansi.Fore.RED, fore_stack)
    [...]

Fore = ForeContexts()
```

As I'm rather new to Github collaboration, could you help me on how to propose this? Should I make a seperate pull request? Thanks.

Moreover, while writing this comment, I thought of yet another way to solve colorama's " + " syntax; why not make *functions* instead of context managers for every fore/back/style? It could then be used as follows:

```python
print(Fore.RED('This message is red.'))
```

The implementation would be largely the same as the context manager implementation I proposed in this comment.

akdor1154:
as a user, I normally write something like this whenever i use colorama, it feels much nicer than manually managing state myself. I'd love to see this included in the library by default, in either of the above forms.

FelixSchwarz:
I also like the ability to let the library handle the display state (so I can't mess up). The [plumbum](https://plumbum.readthedocs.io/en/latest/) library also has this feature: [safe style manipulations](https://plumbum.readthedocs.io/en/latest/colors.html#safe-manipulation)

- What is the next step to get this merged?
- If merging is unlikely/complicated, maybe we should create gist so colorama users can use a colorama-based context manager?

tartley:

The above is a standard cut-and-paste to everyone. For this PR specifically: I haven't looked at it yet, but to judge by skimming the conversation above, it's possible that users who want good functionality for generating ANSI codes for colors, etc, should be using a library like blessings or termcolor that are explicitly dedicated to that. Colorama's reason for existence is to convert those ANSI codes into windows compatible API calls. I foolishly added the 'Fore.RED' shortcuts for creating ANSI codes as a convenience, but long ago wished I hadn't. We should do one thing and do it well, and tell users to use blessings for generating ANSI.  So I'm not yet 100% sure on this PR, but maybe am misunderstanding and am happy to discuss.

tartley:
I'm tremendously grateful to the folks who have contributed to this PR, but for the reasons outlined in my last comment above, I'm going to close this PR without merging, which makes me feel awfully guilty, soreee!. Colorama's support for generating color codes was always meant to be rudimentary, just to check it's working and do something basic - the main point is to make those same codes also work on Windows. If you want something fancy, use a different library to generate the color codes, and use Colorama to make them work on Windows.

PR #163, opened 2018-04-16
Add support for PyCharm

### LINKED ISSUE #107: Can not display color in PyCharm using default init()

veleek:
Fixes #107 
Replaces the abandoned #108

### DIFF
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -18,7 +18,8 @@ def is_stream_closed(stream):
 
 
 def is_a_tty(stream):
-    return hasattr(stream, 'isatty') and stream.isatty()
+    is_pycharm = 'PYCHARM_HOSTED' in os.environ
+    return (hasattr(stream, 'isatty') and stream.isatty()) or is_pycharm
 
 
 class StreamWrapper(object):
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -18,7 +18,8 @@ def is_stream_closed(stream):
 
 
 def is_a_tty(stream):
-    if 'PYCHARM_HOSTED' in os.environ: return True
+    if 'PYCHARM_HOSTED' in os.environ: 
+        return True
     return (hasattr(stream, 'isatty') and stream.isatty())
 
 
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -18,8 +18,8 @@ def is_stream_closed(stream):
 
 
 def is_a_tty(stream):
-    is_pycharm = 'PYCHARM_HOSTED' in os.environ
-    return (hasattr(stream, 'isatty') and stream.isatty()) or is_pycharm
+    if 'PYCHARM_HOSTED' in os.environ: return True
+    return (hasattr(stream, 'isatty') and stream.isatty())
 
 
 class StreamWrapper(object):


veleek:
This change is incredibly simple, and I think it's mostly the intended behavior. PyCharm behaves like a TTY (at least for colorama's intents) so this isolates the changes to one place and in the future if there need to be additional platforms/apps added this would be a reasonable place to do so.

colorama/ansitowin32.py
```diff
@@ -18,7 +18,8 @@ def is_stream_closed(stream):
 
 
 def is_a_tty(stream):
-    return hasattr(stream, 'isatty') and stream.isatty()
+    is_pycharm = 'PYCHARM_HOSTED' in os.environ
```
wiggin15: I think this will look better if we do:
`if 'PYCHARM_HOSTED' in os.environ: return True` and then leave the original return as it was. This is not important, though - what do you think?

wiggin15:
Looks like the request to support PyCharm is quite popular so let's pull it in. See my comment on the commit.

tartley:
@veleek Hi thanks for the input. Looks good to me but I'm going to let @wiggin15 push the button - I don't use Windows any longer so he'll have to support it going forwards.

veleek:
@wiggin15 Sounds good to me.  Made the change and just pushed it.

wiggin15:
Oops. When I wrote `if 'PYCHARM_HOSTED' in os.environ: return True` in one line, I meant it indented, i.e.
```python
if 'PYCHARM_HOSTED' in os.environ:
    return True
```
The one-line works but doesn't look very "pythonic". Sorry for the hassle.

veleek:
:) I'm pretty new to python, so I'm not familiar with the standard styles.

Fixed!

wiggin15:
Thanks for the pull request! This will be included in the next release of colorama (I'm not sure when it will be released).

Delgan:
Hi @veleek and @wiggin15.

Does this change implies that **every** stream checked using `colorama.is_a_tty()` or wrapped using `AnsiToWin32()` will be considered as colorable?

From what I guess looking at the source code, I'm afraid that using `is_a_tty()` with file stream for example will wrongly return `True` on PyCharm, which means we could not use `AnsiToWin32()` with arbitrary stream (calling it with a non-tty stream would wrap it anyway).

I see two possible solutions:
* In `is_a_tty()`, check that `PYCHARM_HOSTED` exists **and** check that the stream is `stderr` or `stdout`
* Move the `PYCHARM_HOSTED` check to the `init()` function

Issue #36, opened 2015-02-16
Invalid no-ASCII unicode string handling


tartley:
Migrated from https://code.google.com/p/colorama/issues/detail?id=21
Reported by av1024, Feb 23, 2011

What steps will reproduce the problem?
1. import and initalize colorama with deafults
2. print u"Some non-ASCII text ТЕСТ Русского"

What is the expected output?
Some non-ASCII text ТЕСТ Русского

What do you see instead?
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-5: ordinal not in range(128)

What version of the product are you using? On what operating system?
Python 2.7 (x32)
Windows 7 x64 Untimate (with Eng/Rus locales)

Please provide any additional information below.
Looks like wrapped write method hoes not inherit/use original stdout encoding. Possible fixes are (two ways):

A. use sys.setdefaultencoding(Your-Console-OEM-Encoding) # Wrong way IMHO. I don't know the simple method to determine right console mode (ANSI/OEM) and OEM encoding except reading 'stdout.encoding' property

B. Patch ansitowin32 to force-encode unicode output before .write:

--- D:\lg\py\colorama-0.1.18\colorama\ansitowin32.py    Tue May 18 14:43:54 2010
+++ ansitowin32.py  Wed Feb 23 19:10:40 2011
@@ -144,7 +144,10 @@
 
     def write_plain_text(self, text, start, end):
         if start < end:
-            self.wrapped.write(text[start:end])
+            if isinstance(text, unicode):
+                self.wrapped.write(text[start:end].encode(self.wrapped.encoding))
+            else:
+                self.wrapped.write(text[start:end])
             self.wrapped.flush()

tartley:
A possibly-easier-to-reproduce test case:

A better string is a single i with acute accent like 'í'

>>> import colorama
>>> colorama.init()
>>> s = u'í'
>>> print repr(s)
u'\xed'
>>> print s
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python26\lib\site-packages\colorama-0.3.1-py2.6.egg\colorama\ansitowi
n32.py", line 35, in write
    self.__convertor.write(text)
  File "C:\Python26\lib\site-packages\colorama-0.3.1-py2.6.egg\colorama\ansitowi
n32.py", line 116, in write
    self.write_and_convert(text)
  File "C:\Python26\lib\site-packages\colorama-0.3.1-py2.6.egg\colorama\ansitowi
[...]

PR #167, opened 2018-06-01
handle unicode strings properly 

jodok:
https://github.com/tartley/colorama/issues/36
https://github.com/Azure/azure-cli/issues/6408


wiggin15:
This change is not compatible with Python 3 (see Travis build). Do you think you can fix it? Thanks

Delgan:
Hi @jodok 

Are you sure the issue is from `colorama` side?

Encoding in Python is notoriously known to be **hard**.

I this the problem may come from environment or terminal.

I could not reproduce your issue, but as you linked to #36, I tried to print `s = u'í'` with and without calling `colorama.init()`.
In both case, the unicode `í` character was properly displayed.
Then I tried with your patch, and when printing `í` ater `colorama.init()`, it was replaced by an empty string `''`. 😕 

I guess my Windows terminal is correctly configured to handle unicode well, I would be disappointed not to be able to enjoy it with `colorama`.

jodok:
my root cause is this issue: https://github.com/Azure/azure-cli/issues/6408 - and there the issue only occurs if it is called in an non-interactive session (e.g. from jenkins).

wiggin15:
Hi @jodok . The fixup for Python 3 looks good but I'm not sure about the solution. If I understand correctly, this change will skip and ignore non-ascii character, which may be a problem, for two reasons:
* The underlying terminal may support non-ascii characters. In this case we can write them and don't need to ignore them.
* This can cause strange output - e.g. when the user prints "Bokmål" (or, say, `u'Bokm\xe5l'`) we will write "Bokml" which doesn't look right.

Perhaps we should consider trying to encode with `self.stream.encoding` instead of ascii?
(by the way, I am also unable to reproduce this reliably)

tartley:

Issue #134, opened 2017-06-21
Add support for ANSI hyperlinks


ssbarnea:
See https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda for details as is already implemented by some major terminals.

wiggin15:
Hi. Thanks for the suggestion. colorama is intended to interpret ANSI codes and convert them so they work in both Linux and Windows. In the list of terminals in the link you provided, I don't see any Windows-based terminals (cmd, PowerShell). If we can't convert the ANSI codes to Windows on Windows machines, there is no point in adding the codes to colorama.. See also the same reasoning in https://github.com/tartley/colorama/issues/38#issuecomment-128834060

stuaxo:
I opened an issue against the windows console to support it

https://github.com/Microsoft/console/issues/204

mixmastamyk:
@ssbarnea  Colorama is not focused on this kind of thing.

However, I made a convenience function in my utility module you might like:
https://github.com/mixmastamyk/console/blob/master/console/utils.py#L54

ssbarnea:
@mixmastamyk I already have a working function https://github.com/pycontribs/gri/blob/master/gri/__main__.py#L31-L32 that was not the problem.

The problem is that I wanted to make it easy for anyone writing python code to use it, and colorama is the probably the best place to implement that functionality.

One way or another colorama is an ANSI processing library, that is an ansi extension. We could easily document that this works only on some terminal.

We already have a rough time dealing with dependencies, creating another module for a one-lines is a very bad idea from the maintenance point of view. 

I personally do not see any issue extending functionality, especially if we add a notice regarding support..

mixmastamyk:
@ssbarnea Well, colorama is not an ansi-utility-library per se.

It's really an "enable your existing ansi-utility-library to work on Windows" library.  One that has served its purpose now that MS is finally getting it together on that front.  Probably now in maintenance mode too.

There are other libraries that might make more sense to add this to.

Issue #122, opened 2017-03-06
Style.ITALIC


arsari:
Does colorama.py have italic style?  Does it considered to be added?

wiggin15:
Hi there.
colorama is meant to work on both Linux and Windows and convert supported ANSI styles to Windows code where possible. While it is possible to produce italic style on some Unix terminals like this:

    print '\x1b[3mThis is in italic!\x1b[0m'

Windows doesn't have a matching console attribute to produce a similar effect ([MSDN](https://msdn.microsoft.com/en-us/library/windows/desktop/ms682088(v=vs.85).aspx#_win32_character_attributes)). Therefore there are no plans to add this style to colorama. If you want to support Unix terminals only, you can always manually add the ANSI codes like above.
See also the issue about other possible ANSI styles that we could not add: 
https://github.com/tartley/colorama/issues/38#issuecomment-128834060

rspencer01:
According to the README, Windows doesn't support DIM either, yet that is included in colorama.  I think this is reasonable to ask to include this, just with a silent fail for Windows.

wiggin15:
You're right about DIM being in the code but not supported. I'm still not sure about adding all the other styles though. There are a lot:

```
    DIM           = 2   # not supported on Windows
    ITALIC        = 3   # not supported on Windows
    UNDERLINE     = 4   # not supported on Windows
    BLINK         = 5   # slow blink. not supported on Windows and not widely supported on Linux
    RBLINK        = 6   # rapid blink. not supported on Windows and not widely supported on Linux
    REVERSEVID    = 7   # not supported on Windows
    CONCEAL       = 8   # not supported on Windows and not widely supported on Linux
    STRIKETHROUGH = 9   # not supported on Windows and not widely supported on Linux
    FRAKTUR       = 20  # not supported on Windows and not widely supported on Linux
```

Adding or not adding the codes is really @tartley's call.
Note that adding the definitions will not really have an effect except for letting users write code like this:
```
print(colorama.Style.ITALIC)
```
instead of printing the codes directly without colorama's help, like this:
```
print('\x1b[3m')
```
or maybe even like this
```
ITALIC = 3
print colorama.ansi.code_to_chars(ITALIC)
```

In the past, @tartley mentioned that if you only want to have convenience definition for the different ANSI codes, you should look at other packages like `termcolor` or `ansicolors`.

tartley:
[...]

Issue #38, opened 2015-02-16
Add support for UNDERLINED, BLINK, REVERSED (with patch!)


tartley:
Migrated from https://code.google.com/p/colorama/issues/detail?id=39
Reported by irmendejong, Dec 7, 2012

I wrote some monkeypatching code to add a few more ansi-styles to colorama.
Unfortunately the win32 style for underlined and reversevideo don't seem to work at all (even though there's constants defined for them). I am simulating the reversevideo one but the underlined one is just not working.

tartley:
diff -r 4ccdb5498413 colorama/ansi.py
--- a/colorama/ansi.py	Sun May 27 10:09:37 2012 +0100
+++ b/colorama/ansi.py	Tue Dec 11 01:23:44 2012 +0100
@@ -38,10 +38,13 @@
     RESET   = 49
 
 class AnsiStyle:
-    BRIGHT    = 1
-    DIM       = 2
-    NORMAL    = 22
-    RESET_ALL = 0
+    BRIGHT     = 1
+    DIM        = 2
+    UNDERLINED = 4
+    BLINK      = 5
+    REVERSEVID = 7
+    NORMAL     = 22
+    RESET_ALL  = 0
 
 Fore = AnsiCodes( AnsiFore )
 Back = AnsiCodes( AnsiBack )
diff -r 4ccdb5498413 colorama/ansitowin32.py
--- a/colorama/ansitowin32.py	Sun May 27 10:09:37 2012 +0100
+++ b/colorama/ansitowin32.py	Tue Dec 11 01:23:44 2012 +0100
@@ -89,6 +89,7 @@
                 AnsiStyle.BRIGHT: (winterm.style, WinStyle.BRIGHT),
                 AnsiStyle.DIM: (winterm.style, WinStyle.NORMAL),
                 AnsiStyle.NORMAL: (winterm.style, WinStyle.NORMAL),
+                AnsiStyle.REVERSEVID: (winterm.style_reverse_vid, ),
                 AnsiFore.BLACK: (winterm.fore, WinColor.BLACK),
                 AnsiFore.RED: (winterm.fore, WinColor.RED),
                 AnsiFore.GREEN: (winterm.fore, WinColor.GREEN),
diff -r 4ccdb5498413 colorama/tests/ansi_test.py
--- a/colorama/tests/ansi_test.py	Sun May 27 10:09:37 2012 +0100
[...]

PR #168, opened 2018-06-05
Adds link method for adding hyperlinks

ssbarnea:
Fixes: #134

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ dist
 tags
 virtualenv
 build
+.tox
 
 # PyCharm
 .idea
--- a/README.rst
+++ b/README.rst
@@ -146,6 +146,18 @@ Available formatting constants are::
 ``Style.RESET_ALL`` resets foreground, background, and brightness. Colorama will
 perform this reset automatically on program exit.
 
+Links
+-----
+
+Links are a new (2017) ANSI extension that mimics the HTML <a> blocks. You should
+be using it with care as many terminals are not supporting it yet.
+
+.. code-block:: python
+
+    from colorama import init, link
+
+    init()
+    print(link('http://example.com', 'click me'))
 
 Cursor Positioning
 ------------------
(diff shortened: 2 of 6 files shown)


ssbarnea:
@egmontkob

wiggin15:
Hi. Thanks for the PR. The problem I see with this is that there is no comparable feature (that I know of) on Windows terminals (e.g. "cmd"), and colorama is all about making the ANSI codes support both Windows and Unix platforms (more than convenience for creating and printing the codes). 
See also https://github.com/tartley/colorama/issues/38#issuecomment-128834060.

skyem123:
It's probably possible to imitate this feature on Windows, but I think that'd be like reinventing curses (and I don't know if you can get mouse input while also having a scrollback)?

ssbarnea:
@wiggin15 The fact that one specific terminal does no supports a feature is not a good enough reason for not implementing a feature at all. I am aware about at least 4 different terminals used on windows: old (bad) cmd, powershell, new bash terminal from linux subsystem (apparently different than cmd)  and putty. 

I stopped using Windows machines years ago so it may be quite hard to test especially due to their various versions. Also, it would not be hard to add some conditionals for disabling this functionality on win32 but I am afraid it is impossible to determine if current terminal supports it or not. We can only assume that is up to the colorama library user to decide when to use this feature or not.

So please let me know what else can be do to implement this feature in colorama?

wiggin15:
This isn't about one specific terminal.
The purpose of colorama, as written in the README:

> Makes ANSI escape character sequences (for producing colored terminal text and cursor positioning) work under MS Windows.

We do this by converting ANSI codes to Win32 API calls when we see them written to stdout/stderr. We try to support as many ANSI codes as possible (not just colors), but only if they _can_ be converted to Windows API for Windows terminals (cmd and the others). When this is not possible, we don't add support for the codes.
This is true for other ANSI codes that we couldn't convert, like Underline, Blink (#38), Italic (#122), etc.
Unfortunately I don't think there is a Windows function to add a link to a Windows console.

I appreciate the effort and wanting to contribute, and I know it is frustrating that we can't add all available codes, but this is a decision we made long ago.

Check out [blessings](https://pypi.org/project/blessings) or other packages that are intended to provide convenience methods for printing ANSI codes, and should work with colorama.

PR #175, opened 2018-06-14
Replace calls to dict() with dict literals

jdufresne:
Literals are more idiomatic Python and are always (very slightly) faster.


wiggin15:
Your pull requests are great, thank you! This one however is more of a styling preference AFAIK, and I'm inclined to favor the current style. The "dict()" call supports args/kwargs, and using parameters (as in `dict(a='b')`) instead of strings looks better IMHO.

jdufresne:
I agree there is a bit of a style choice here. If you prefer the current style, no problem. I don't mind closing the PR.

PR #183, opened 2018-08-21
Include tests in source distributions

sbraz:
Hi,
We at Gentoo rely on PyPI tarballs to build docs and run tests, could you include them in future releases so that we don't have to use GitHub's tarballs?

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,5 @@
 include LICENSE.txt CHANGELOG.rst
 recursive-include demos *.py *.bat *.sh
+graft tests
+global-exclude __pycache__
+global-exclude *.py[co]


tartley:

sbraz:
Hi @tartley. If your PR works, it means tests are included so my PR would not be needed :)

tartley:
Understood, thanks for the clarification. I'll double check when  I'm not meant to be doing my real job... :-)

sbraz:
No worries, it's not a massive problem. We (Gentoo) simply use GitHub releases instead of PyPI packages for now. Howevern, it is good practice to include tests in sdists if they aren't too large. I know Debian also relies on this to test packages.

mtelka:
+1 for including tests in sdist (including the `tox.ini` file).  OpenIndiana have similar workflow like Gentoo (and others mentioned above) so tests in sdist would make the world better.  Thank you.

PR #186, opened 2018-09-07
added contextlib magic methods to ansitowin32.StreamWrapper

### LINKED ISSUE #185: colorama.init(autoreset=True) breaks argparse
Consider the following snippet:

```python3
#!/usr/bin/env python3
import argparse
import colorama
#colorama.init(autoreset=True)


if __name__ == '__main__':
    parser = argparse.ArgumentParser()
    parser.add_argument('-o', '--output', type=argparse.FileType('w'), default='-')
    args = parser.parse_args()
    with args.output as fp:
        fp.write('hello world\n')
```

When running, it prints `hello world` as supposed:

```sh
$ ./spam.py
hello world
```

However, uncommenting `colorama.init(autoreset=True)` breaks the script when entering the `with` context:

```sh
$ ./spam.py
Traceback (most recent call last):
  File "spam.py", line 10, in <module>
    with args.output as fp:
AttributeError: __enter__
```

I uncovered this issue in a setup where another package pulls `colorama` and calls `colorama.init(autoreset=True)` under the hood, with a side-effect of breaking my own code. The reason is `StreamWrapper` wraps `io.TextIOWrapper`, but doesn't provide access to the `__enter__` and `__exit__` magic methods.

May I suggest adding `__enter__` and `__exit__` magic methods to `StreamWrapper`?

```python3
class StreamWrapper(object):
    '''
    Wraps a stream (such as stdout), acting as a transparent proxy for all
    attribute access apart from method 'write()', which is delegated to our
    Converter instance.
    '''

    def __init__(self, wrapped, converter):
        # double-underscore everything to prevent clashes with names of
        # attributes on the wrapped stream object.
        self.__wrapped = wrapped
        self.__convertor = converter

    def __getattr__(self, name):

    def __enter__(self, *args, **kwargs):
        return self.__wrapped.__enter__(*args, **kwargs)

    def __exit__(self, *args, **kwargs):
        return self.__wrapped.__exit__(*args, **kwargs)

    def write(self, text):
        self.__convertor.write(text)
```
Right now, it's either that (monkeypatch `StreamWrapper` with the above class if importable) or checking for private attributes in the package:

```python3
out = args.output._StreamWrapper__wrapped if hasattr(args.output, '_StreamWrapper__wrapped') else args.output
with out as fp:
    fp.write('hello world\n')
```

hoefling:
Fixes #185.

### DIFF
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -39,6 +39,9 @@ class StreamWrapper(object):
         return getattr(self.__wrapped, name)
 
     def __enter__(self, *args, **kwargs):
+        # special method lookup bypasses __getattr__/__getattribute__, see
+        # https://stackoverflow.com/questions/12632894/why-doesnt-getattr-work-with-exit
+        # thus, contextlib magic methods are not proxied via __getattr__
         return self.__wrapped.__enter__(*args, **kwargs)
 
     def __exit__(self, *args, **kwargs):
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -38,6 +38,12 @@ class StreamWrapper(object):
     def __getattr__(self, name):
         return getattr(self.__wrapped, name)
 
+    def __enter__(self, *args, **kwargs):
+        return self.__wrapped.__enter__(*args, **kwargs)
+
+    def __exit__(self, *args, **kwargs):
+        return self.__wrapped.__exit__(*args, **kwargs)
+
     def write(self, text):
         self.__convertor.write(text)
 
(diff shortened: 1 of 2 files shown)


wiggin15:
`StreamWrapper` has `__getattr__` which should return the correct attributes for `__enter__` and `__exit__` from the wrapped stream `self.__wrapped`. Theoretically this should Just Work but it doesn't. I saw this that answer that sort of explains why it doesn't:
https://stackoverflow.com/questions/12632894
@hoefling  do you mind adding this link in a comment in the code so it will be clear why these methods are needed? Thanks.

wiggin15:
Looks good. Thanks.

PR #187, opened 2018-09-09
Implements Color Incrementation 

osilkin98:
This is a simply function added to `ansi.py` which allows users to increment colors to allow users to create rainbow-like outputs. This would be beneficial as it would allow users to implement features to distinguish mundane outputs from one another. 

Personally I'm amazed that this wasn't already implemented as color codes should be treated as numbers because they fundamentally are, if you strip away the extra ANSI codes.

### DIFF
--- a/colorama/ansi.py
+++ b/colorama/ansi.py
@@ -22,6 +22,48 @@ def clear_line(mode=2):
     return CSI + str(mode) + 'K'
 
 
+# This function needs to be able to take ANY given ANSI Color String and return
+# It's value as the next incremented color code.
+def increment_color(code):
+    """ Takes a Given Color Code Defined in a Subclass of AnsiCodes,
+    and returns the next modulating non-reseting color.
+
+    :param code: ANSI Color Constant defined in either Fore, Back, or Style.
+     These Color Constants come in the form '\033['+ X + 'm', where X is the actual integer that we want.
+    :type code: str
+    :return: Another ANSI Color Constant defined in the corresponding subclasses
+    :rtype: str
+    """
+    # Extract the number from the code, this strips all the strings away from the number,
+    # And allows us to simply convert from string to integer directly
+    color_num = int(code.lstrip(CSI).rstrip('m'))
+
+    # Fore Integers are defined 30-37, 90-97. 39 is a reset value, and 38 is a blank
+    # Back Integers are defined 40-47, 100-107, 49 is a reset value, and 48 is also a blank
+    # Style Integers are defined as 1,2 & 22. 0 is a reset value.
+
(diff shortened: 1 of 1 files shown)


colorama/ansi.py
```diff
@@ -22,6 +22,48 @@ def clear_line(mode=2):
...
+    # Style Integers are defined as 1,2 & 22. 0 is a reset value.
+
+    # We want the Style, Fore, and Back integers to stay in their respective class,
+    # Which eliminates the possibility of using a circular buffer.
+
+    # Instead we have to use this spaghetti code logic in order to handle the fringe cases
+    # Ensuring that we're incrementing when needed and decrementing to the correct positions.
+
+    # Handle the fringe cases for the Fore and Back ANSI codes to jump to the non-standard codes
+    color_num = color_num + 53 if color_num == 47 or color_num == 37 else\
```
osilkin98: There's no easy way to handle the incrementation with a shorthand logic, circular buffers fail as they still require some sense of what number is & isn't in a given set. 
Therefore the only simple solution is to use this long spaghetti logic.

wiggin15:
I'm afraid I don't see a clear, _general_ use case for this, and the function is too littered with magic numbers and specifics for _how_ to generate the rainbow effect. IMO this sort of code belongs in the application(s) using this package, instead of in the package itself.

btw, demo01.py also outputs a sort of rainbow of colors, but goes about it very differently.

osilkin98:
> I'm afraid I don't see a clear, _general_ use case for this
The use case is when you want to increment a color in value, as if it were a number

> the function is too littered with magic numbers 
They're not magic, but I wasn't able to put end of line comments for the assignment statement due to 
the fact that I'd escape it to the next to avoid a long statement on one line

> the function is too littered with specifics for how to generate the rainbow effect
That's not true at all, the comments explain how the numbers bounce back and forth for the 3 different classes that define color constants; with no reference to any sort of "rainbow effect". Secondly, if verbosity is an issue for you, I can always just reduce the amount of comments.

colorama/ansi.py
```diff
@@ -22,6 +22,48 @@ def clear_line(mode=2):
...
+                color_num - 67 if color_num == 107 or color_num == 97 else\
+                color_num + 1 if color_num >= 30 and color_num != (39 and 49) else\
+                1 if color_num == 22 else\
+                2 if color_num == 1 else\
+                22 if color_num == 2 else\
+                1 if color_num == 0 else color_num + 51
```
jcs090218: I get the idea of what these operations does, yet is there a better way to implement this? I mean on the aspect of readability and all those magic numbers.

tartley:

PR #189, opened 2018-09-15
Use the recommended test command in tox.ini

jdufresne:
Running the command `tox` now tests in the same way as Travis CI and as
recommended in the docs. Makes it simpler to recreate an identical test
environment locally.


wiggin15:
Thanks!

PR #191, opened 2018-09-15
Remove unnecessary distutils fallback

jdufresne:
The distutils fallback isn't necessary. setuptools should be available on all modern, supported Pythons.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,10 +5,7 @@ from __future__ import with_statement
 
 import os
 import re
-try:
-    from setuptools import setup
-except ImportError:
-    from distutils.core import setup
+from setuptools import setup
 
 
 NAME = 'colorama'


wiggin15:
setuptools is still an external package and not something in the Python standard library, so it's not too hard to find or create an environment without it. I know a lot of other packages have this fallback. There's no harm in having it.

jdufresne:
I'm curious, where are you encountering Python environments without setuptools? From the [Python Packaging User Guide](https://packaging.python.org/guides/installing-using-linux-tools/), setuptools and pip are included with python.org Python downloads. For Linux distributions, setuptools is usually installed by the system package manager. As `setup.py` is using setuptools specific features, always using it would make for a more predictable and consistent installation experience across environments.

wiggin15:
I can think of a few examples where you can encounter a "bare" Python:
* Older installations of Python from python.org (2.7..) don't come with setuptools.
* on Linux systems that come with Python 3, you can install Python 2 (or the other way around) using the operating system package manager (yum/apt) but not install the corresponding python-setuptools.
* `virtualenv --no-setuptools`
* Unix systems where you install Python from their own package format (bff/pkg/rpm - without a package manager)
* compile Python from source :) or projects that compile a special version of Python and distribute it in their own package (e.g. relocatable-python).

jdufresne:
I see. Thanks for the informative answer.

Issue #196, opened 2018-10-13
PR 164 causes hang when autoreset=True


bbayles:
Updating to 0.4.0 caused an application I work on to stop working. Investigating, I found that getting rid of `autoreset=True` fixed the hanging problem, but I need the `autoreset` behavior.

Working through the changes since 0.3.9, I found that reverting a3eabd264365158f9d213a7b0a512c6955f0d281 fixes the problem. This the merge for PR #164.

I've not nailed down why this affects me - I'm not using PyCharm - but I expect you'll hear from others as this update proliferates. If I can pin it down I'll submit a PR.

bbayles:
Update: more specifically, it's 2057f036f23a8113587f9cd7bf558594444a9caf that causes the hang. Keeping the PyCharm behavior from 3579cdec48d86cc8f588a3c8b5c990f1e0118c95 seems fine; it's the moving of the function to a method that causes the problem.

wiggin15:
Fixed in #198.

wiggin15:
Uploaded v0.4.1 to PyPI. Thanks.

PR #197, opened 2018-10-13
Rename .closed property to avoid recursion

bbayles:
Re: issue #196, this PR changes one aspect of #164 that was causing me problems.

When using `autoreset=True`, [this check](https://github.com/tartley/colorama/blob/c11da6f6004bba0fabb2f7a68a06f53baeeb6b74/colorama/ansitowin32.py#L165) seems to happen over and over. [This method](https://github.com/tartley/colorama/blob/c11da6f6004bba0fabb2f7a68a06f53baeeb6b74/colorama/ansitowin32.py#L53) is calling itself, since `StreamWrapper` can be the stream.

Adding a print statement [here](https://github.com/tartley/colorama/blob/c11da6f6004bba0fabb2f7a68a06f53baeeb6b74/colorama/ansitowin32.py#L52):
```
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...
Checking closure...

```

So we basically have a name collision. In order to avoid that I've renamed the `StreamWrapper.closed` property to `.is_closed`. I'm happy to name it something else, or turn it back into a method instead of a property.

### DIFF
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -48,7 +48,8 @@ class StreamWrapper(object):
         return (hasattr(stream, 'isatty') and stream.isatty())
 
     @property
-    def closed(self):
+    def is_closed(self):
+        print('Checking!')
         stream = self.__wrapped
         try:
             return stream.closed
@@ -84,12 +85,12 @@ class AnsiToWin32(object):
 
         # should we strip ANSI sequences from our output?
         if strip is None:
-            strip = conversion_supported or (not self.stream.closed and not self.stream.isatty())
+            strip = conversion_supported or (not self.stream.is_closed and not self.stream.isatty())
         self.strip = strip
 
         # should we should convert ANSI sequences into win32 calls?
         if convert is None:
-            convert = conversion_supported and not self.stream.closed and self.stream.isatty()
+            convert = conversion_supported and not self.stream.is_closed and self.stream.isatty()
         self.convert = convert
 
         # dict of ansi codes to win32 functions and parameters
@@ -165,7 +166,7 @@ class AnsiToWin32(object):
     def reset_all(self):
         if self.convert:
             self.call_win32('m', (0,))
-        elif not self.strip and not self.stream.closed:
+        elif not self.strip and not self.stream.is_closed:
             self.wrapped.write(Style.RESET_ALL)
 
 
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -49,7 +49,6 @@ class StreamWrapper(object):
 
     @property
     def is_closed(self):
-        print('Checking!')
         stream = self.__wrapped
         try:
             return stream.closed
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -48,9 +48,12 @@ class StreamWrapper(object):
         return (hasattr(stream, 'isatty') and stream.isatty())
 
     @property
-    def is_closed(self):
+    def closed(self):
         stream = self.__wrapped
-        return not hasattr(stream, 'closed') or stream.closed
+        try:
+            return stream.closed
+        except AttributeError:
+            return True
 
 
 class AnsiToWin32(object):
@@ -81,12 +84,12 @@ class AnsiToWin32(object):
 
         # should we strip ANSI sequences from our output?
         if strip is None:
-            strip = conversion_supported or (not self.stream.is_closed and not self.stream.isatty())
+            strip = conversion_supported or (not self.stream.closed and not self.stream.isatty())
         self.strip = strip
 
         # should we should convert ANSI sequences into win32 calls?
         if convert is None:
-            convert = conversion_supported and not self.stream.is_closed and self.stream.isatty()
+            convert = conversion_supported and not self.stream.closed and self.stream.isatty()
         self.convert = convert
 
         # dict of ansi codes to win32 functions and parameters
@@ -162,7 +165,7 @@ class AnsiToWin32(object):
     def reset_all(self):
         if self.convert:
             self.call_win32('m', (0,))
-        elif not self.strip and not self.stream.is_closed:
+        elif not self.strip and not self.stream.closed:
             self.wrapped.write(Style.RESET_ALL)
 
 
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -48,7 +48,7 @@ class StreamWrapper(object):
         return (hasattr(stream, 'isatty') and stream.isatty())
 
     @property
-    def closed(self):
+    def is_closed(self):
         stream = self.__wrapped
         return not hasattr(stream, 'closed') or stream.closed
 
@@ -81,12 +81,12 @@ class AnsiToWin32(object):
 
         # should we strip ANSI sequences from our output?
         if strip is None:
-            strip = conversion_supported or (not self.stream.closed and not self.stream.isatty())
+            strip = conversion_supported or (not self.stream.is_closed and not self.stream.isatty())
         self.strip = strip
 
         # should we should convert ANSI sequences into win32 calls?
         if convert is None:
-            convert = conversion_supported and not self.stream.closed and self.stream.isatty()
+            convert = conversion_supported and not self.stream.is_closed and self.stream.isatty()
         self.convert = convert
 
         # dict of ansi codes to win32 functions and parameters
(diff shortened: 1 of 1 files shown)


bbayles:
The problem is that things stack up. If I add a debug statement to the `closed` property on the `master` branch:

```python
    @property
    def closed(self):
        print('is closed?')
        stream = self.__wrapped
        return not hasattr(stream, 'closed') or stream.closed
```

Then do something like this will produce this output:

```
>>> import colorama
... import sys
... 
... colorama.init(autoreset=True)
is closed?
is closed?
is closed?
is closed?
```

Switching to an older version (or my branch here), you get this:

```
>>> import colorama
... import sys
... 
... colorama.init(autoreset=True)
is closed?
is closed?
```

So there are more checks.

Subsequent calls to init (on `master`) produce an explosion of checks - I found this when running unit tests, which exercise the same code over and over. The second call produces 12 checks. The third produces 28. The fourth 60... I think it's 2^n - 4 calls.

On an older verison (or my branch here), it's always two checks.

Delgan:
@bbayles Thanks for the explanations!

So, there is no infinite recursion nor method calling itself (just method calling the same method, but the one of the wrapped stream), right? The issue is caused by too much nested calls, occuring while checking `.closed` in `reset_all()`, which is only called if `autoreset=True`.

To better vizualise what caused these excessive calls after https://github.com/tartley/colorama/commit/2057f036f23a8113587f9cd7bf558594444a9caf, I draw the call stack before and after your fix.

Before your fix, accessing `stream.closed` results in first calling `hasattr(wrapped, 'closed')` and then accessing `wrapped.closed`.

```python
@property
def closed(self):
    stream = self.__wrapped
    return not hasattr(stream, 'closed') or stream.closed
```

But as `hasattr(wrapped, 'closed')` is implemented by calling `getattr(wrapped, 'closed')` (and checking if an error occurs), this causes exploding nested calls as this is equivalent to accessing `wrapped.closed` and hence double checking the nested stream, and this get worse as the number of nested streams increases. The same problem happens while calling `.isatty()`.

For example if `init()` is called three times, the resulting stream is equivalent to `StreamWrapper(StreamWrapper(StreamWrapper(sys.stderr)))`, let's rename it to `A(B(C(sys.stderr)))`.

Before your fix, the call stack would look like this:

```
A.closed
  ├ hasattr(B, 'closed')
  │  └ getattr(B, 'closed')
  │       ├ hasattr(C, 'closed')
  │       │   └ getattr(C, 'closed')
  │       │       ├ hasattr(sys.stderr, 'closed')
  │       │       │   └ getattr(sys.stderr, 'closed')
  │       │       └ sys.stderr.closed
  │       └ C.closed
  │           ├ hasattr(sys.stderr, 'closed')
  │           │   └ getattr(sys.stderr, 'closed')
  │           └ sys.stderr.closed            
  └ B.closed
      ├ hasattr(C, 'closed')
      │   └ getattr(C, 'closed')
      │       ├ hasattr(sys.stderr, 'closed')
      │       │    └ getattr(sys.stderr, 'closed')
      │       └ sys.stderr.closed
      └ C.closed
          ├ hasattr(sys.stderr, 'closed')
          │   └ getattr(sys.stderr, 'closed')
          └ sys.stderr.closed  
```

As your fix removes the `.closed` property, trying to accessing it fallbacks to `__getattr__`.

```python
def __getattr__(self, name):
```

The calls stack looks like this while accessing `.is_closed`:

```
A.is_closed
  ├ hasattr(B, 'closed')
  │   └ getattr(B, 'closed')
  │       └ __getattr__(B, 'closed')
  │           └ getattr(C, 'closed')
  │               └ __getattr__(C, 'closed')
  │                   └ getattr(sys.stderr, 'closed')
  └ B.closed
      └ __getattr__(B, 'closed')
          └ getattr(C, 'closed')
              └ __getattr__(C, 'closed')
                  └ getattr(sys.stderr, 'closed')
```

Much shorter, and prevent the number of calls to exponentially grows! 

I think we can have the best of both worlds, that is keep the linear number of calls while accessing `.closed`, by simply replacing the `hasattr()` call appropriately.

What do you think of re-implementing the property like this:

```python
@property
def closed(self):
    stream = self.__wrapped
    try:
        return stream.closed
    except AttributeError:
        return True
```

The resulting call stack would be:

```
A.closed
  └ B.closed
      └ C.closed
          └ sys.stderr.closed
```

Even simpler!

The same fix should be done for `.isatty()`. That's way, the `StreamWrapper()` would stay transparently equivalent to the stream it wraps.

bbayles:
The `try`/`except` looks nicer, but I still get exponential behavior when I test your method.

```python
    @property
    def closed(self):
        print('is this closed?')
        stream = self.__wrapped
        try:
            return stream.closed
        except AttributeError:
            return True
```

```
>>> import colorama
... colorama.init(autoreset=True)
is this closed?
is this closed?
is this closed?
is this closed?
>>> colorama.init(autoreset=True)
is this closed?
is this closed?
is this closed?
is this closed?
is this closed?
is this closed?
is this closed?
is this closed?
```

Delgan:
@bbayles It doesn't seem exponential on my computer, more like `2 6 12 20 30 42 56 72 90`.

You should also tries to add the `print()` in the `__getattr__()` under the condition `if name == "closed"` to fully see all recursion calls.

By the way, this makes me wonder... As the `print()` is called from inside the colorama module, does not that risk to interfer?... Because as `sys.stdout` is wrapped by a `StreamWrapper` with `autoreset=True`, calling `print()` will trigger the `.closed` check in `reset_all()`, hence causing excessive calls. 🤔 

I made my tests by adding a global `n` variable in `ansitowin32.py` that I increment within the `.closed` property, and I added a `print_n()` function that I call between two `init()`.

bbayles:
Re: whether it's exponential, I think the number shouldn't be growing at all.

Re: printing; we can reproduce the same issue printing to stderr instead of stdout; I don't think it matters.

My main thought is that the exponential behavior in the currently released version is significant, and will probably cause problems with downstream projects, at least in test suites. It's probably appropriate to apply my fix here (or revert 2057f036f23a8113587f9cd7bf558594444a9caf) and release 0.4.1?

wiggin15:
> By the way, this makes me wonder... As the `print()` is called from inside the colorama module, does not that risk to interfer?

It does. The reason you get this growing number of calls is because you are using "autoreset" and debug-printing from inside the "closed" property. When colorama auto-resets, it checks whether the stream is closed to print the reset sequence. So if every call to "closed" causes a print, it also causes a reset and causes another call to "closed". That would explain the recursion in the first case and the growing prints in after your changes.
In other words, it looks like your debug-printing is what is causing the trouble. There must be another issue in #196, but there isn't enough information there to investigate.

Delgan:
> Re: whether it's exponential, I think the number shouldn't be growing at all.

@bbayles The number has to grown, each subsequent call to `init()` wrap and replace the `sys.stderr`. So, if you wrap an already wrapped stream, accessing `.closed` will need two calls down to the original `sys.stderr` where the `closed` value is retrieved. If you call `init()` again, three calls will be needed, etc. The only reason you may think this number don't grow with your implementation is because you don't add `print()` in the `__getattr__` functon, but the nested calls are here.


> My main thought is that the exponential behavior in the currently released version is significant, and will probably cause problems with downstream projects, at least in test suites. It's probably appropriate to apply my fix here (or revert [2057f03](https://github.com/tartley/colorama/commit/2057f036f23a8113587f9cd7bf558594444a9caf)) and release 0.4.1?

I agree with you, the current code on `master` is prone to exponential calls. My thought is that we can fix it by re-implementing the `.closed` property properly rather than adding a new method.

> There must be another issue in #196, but there isn't enough information there to investigate.

@wiggin15 I think I understood the issue, as explained in my second message, there is indeed a problem with excessive nested calls, even by omitting erroneous calls to `print()`. The callstacks should be quite straightforward to vizualise the issue (at least I thought :grin: ).

bbayles:
Re: issue #196, the exponential calls were causing my hanging problem.

Re: a fix, you're welcome to close this PR and go with some other fix. I'm happy as long as PyPI gets updated promptly to remove the exponential behavior.

Delgan:
@bbayles Yes, fixing this should be done as soon as possible, thanks again for having reported the issue I introduced... 😀 

I implemented an alternative fix solution as discussed, see #198

PR #198, opened 2018-10-15
Prevent exponential number of calls while accessing "closed" / "isatty" (#196)

Delgan:
See #197 for extended discussion.

### DIFF
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -45,12 +45,20 @@ class StreamWrapper(object):
         if 'PYCHARM_HOSTED' in os.environ:
             if stream is not None and (stream is sys.__stdout__ or stream is sys.__stderr__):
                 return True
-        return (hasattr(stream, 'isatty') and stream.isatty())
+        try:
+            stream_isatty = stream.isatty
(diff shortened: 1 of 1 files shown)


bbayles:
This does fix the problem in #196.

wiggin15:
@Delgan  I _think_ I understand [your comment](https://github.com/tartley/colorama/pull/197#issuecomment-429637613) in #197 about the call traces being more complex than needed, but I don't think I follow what the problem in #196 is exactly. If we wrap the streams too many times then this fix will only be a temporary relief and not a real fix, because there will still be too many nesting calls. Won't there..? If I understand correctly the problem was _not_ maximum recursion?

It will help to understand the problem if you can provide a minimal sample code (without modifying colorama) that reproduces this and explain exactly what happens. Thanks.

wiggin15:
Thanks @Delgan . This is a fairly simple change and if it has this big an effect I will merge this. One question, though: if we revert the change in 2057f03 (make `closed` and `isatty` properties), will that make the potential `RecursionError` in this case go away?

wiggin15:
Thanks!

Issue #182, opened 2018-07-04
RecursionError exception if init() called multiple times on Win32


dcleland:
colorama 0.3.9
Windows 7 SP1 (64-bit)
Python 3.6.5 (v3.6.5:f59c0932b4, Mar 28 2018, 17:00:18) [MSC v.1900 64 bit (AMD64)] on win32

If you call init() multiple times you will eventually get a RecursionError exception.

Simple code to reproduce the error. It fails on the 249th iteration

```
#!/usr/bin/python3
from colorama import init
from inspect import getouterframes, currentframe

try:
    for it in range(1, 1001):
        print("Iteration: ", it)
        init()

except Exception:
    one, two, three = sys.exc_info()
    print("{0}\n {1}\n{2}\n".format(one, two, three ))
    

```    
**NOTE**: You will not see the output from the 'except Exception' section, instead you will see

```
  Iteration: Traceback (most recent call last):

  During handling of the above exception, another exception occurred:

```
From a different script that first showed the problem I managed to get the following stack backtrace

```
Iteration: 248
Level is 81,2,3  level  <class 'RecursionError'> === maximum recursion depth exceeded while calling a Python object === <traceback o
bject at 0x0000000003927548> --- 7<----------
Level is 7
Traceback (most recent call last):
  File "W:\XXXXXXXX\test_unsecured_menu_01.py", line 265, in <module>
    test01()
  File "W:\XXXXXXXX\test_unsecured_menu_01.py", line 23, in test01
    print("Level is {}".format( level))   
  File "C:\Python36_64\lib\site-packages\colorama\ansitowin32.py", line 40, in write
    self.__convertor.write(text)
  File "C:\Python36_64\lib\site-packages\colorama\ansitowin32.py", line 141, in write
    self.write_and_convert(text)
  File "C:\Python36_64\lib\site-packages\colorama\ansitowin32.py", line 169, in write_and_convert
    self.write_plain_text(text, cursor, len(text))
  File "C:\Python36_64\lib\site-packages\colorama\ansitowin32.py", line 174, in write_plain_text
    self.wrapped.write(text[start:end])    
  File "C:\Python36_64\lib\site-packages\colorama\ansitowin32.py", line 40, in write
    self.__convertor.write(text)
  File "C:\Python36_64\lib\site-packages\colorama\ansitowin32.py", line 141, in write
    self.write_and_convert(text)
[...]

Issue #145, opened 2017-09-26
deinit() doesn't work if init() was called twice.


gilch:
```Python
>>> import colorama
>>> import sys
>>> id(sys.stdout)
2198406407512
>>> colorama.init()
>>> id(sys.stdout)
2198419631912
>>> colorama.deinit()
>>> id(sys.stdout)  # back to the original, as expected
2198406407512
>>> colorama.init()
>>> colorama.init()
>>> colorama.deinit()
>>> id(sys.stdout)  # What?
2198419494224
```
I think `init()` should probably either raise an error, or no-op if colorama was already initialized.
But in any case, it should not break `deinit` by losing the original `stdin`.

scrunch077:
I oringinally thought that this problem could be solved by just running `deinit()` again (as `init()` was called twice therefore to go back would require two `deinit()` calls) to get back to the original stdout.

However this does not seem to be the case when I started investigating, I found out that the second `deinit()` does not have any affect on the output of stdout, however the first `deinit()` seems to have some effect and lets you go back to the stdout output after the first `init()` was called.
```python
>>> import colorama
>>> import sys
>>> id(sys.stdout)  # The original stdout output
56145200
>>> colorama.init()
>>> id(sys.stdout)  # The stdout output after init()
94884112
>>> colorama.deinit()
>>> id(sys.stdout)  # Should be back to the original stdout output
56145200  # stdout goes back

>>> colorama.init()
>>> id(sys.stdout)  # The stdout output after init()
95230320
>>> colorama.init()  # init() for the second time
>>> id(sys.stdout)  # The stdout output after the second init()
95230544
>>> colorama.deinit()  # deinit() once
>>> id(sys.stdout)  # stdout output after first deinit()
95230320  # stdout goes back to stdout after first deinit()
>>> colorama.deinit() # deinit() again (second time), should go back to original stdout ouput
>>> id(sys.stdout)  # stdout output after second deinit()
95230320  # stdout output is the same after the second deinit() as after the first deinit()
```
For the curious, the same effect occurs when more calls to `init()` are made without calling `deinit()`. Calling `init()` 3 times in a row and then calling `deinit()` 3 times makes stdout print out the same result as after the first `init()` this implies that multiple `init()` calls have an effect on stderr while multiple `deinit()` calls have no effect on stderr after the first `deinit()` call was made.

I hope this helps, I think an error needs to be raised when init() is called again before deinit() is called to prevent this problem happening.

tartley:
[...]

Issue #140, opened 2017-08-11
Infinite recurse on windows


ovidiub13:
When running on Windows, with `init(convert=True)` I get this.
Tested on Windows 8.1 and Windows 10.
When used on Linux, I get no colors.

```
Traceback (most recent call last):
File "C:\jenkins\workspace\main.py", line 77, in <module>
  ts.append(testSuite.runTestSuite())
File "C:\jenkins\workspace\TTP_TestSuite.py", line 67, in runTestSuite
  currentTest.runTest()
File "C:\jenkins\workspace\TTP_Test.py", line 201, in runTest
  self.output.info("Running test: %s" % self.name, 1)
File "C:\jenkins\workspace\output.py", line 29, in info
  self.writeln(data, verbose, front=(Style.BRIGHT +  Fore.CYAN))
File "C:\jenkins\workspace\output.py", line 18, in writeln
  self.write(data, verbose, front, error, True)
File "C:\jenkins\workspace\output.py", line 26, in write
  print(data, end=end, file=out)
File "C:\Program Files\Python36\lib\site-packages\colorama\ansitowin32.py", line 40, in write
  self.__convertor.write(text)
File "C:\Program Files\Python36\lib\site-packages\colorama\ansitowin32.py", line 141, in write
  self.write_and_convert(text)
File "C:\Program Files\Python36\lib\site-packages\colorama\ansitowin32.py", line 166, in write_and_convert
  self.write_plain_text(text, cursor, start)
File "C:\Program Files\Python36\lib\site-packages\colorama\ansitowin32.py", line 174, in write_plain_text
  self.wrapped.write(text[start:end])
File "C:\Program Files\Python36\lib\site-packages\colorama\ansitowin32.py", line 40, in write
  self.__convertor.write(text)
File "C:\Program Files\Python36\lib\site-packages\colorama\ansitowin32.py", line 141, in write
  self.write_and_convert(text)
File "C:\Program Files\Python36\lib\site-packages\colorama\ansitowin32.py", line 169, in write_and_convert
  self.write_plain_text(text, cursor, len(text))
File "C:\Program Files\Python36\lib\site-packages\colorama\ansitowin32.py", line 174, in write_plain_text
  self.wrapped.write(text[start:end])
File "C:\Program Files\Python36\lib\site-packages\colorama\ansitowin32.py", line 40, in write
  self.__convertor.write(text)
File "C:\Program Files\Python36\lib\site-packages\colorama\ansitowin32.py", line 141, in write
[...]

Issue #128, opened 2017-04-20
Provide `isinitialized` function


bdarnell:
In tornadoweb/tornado#2019, we need a way to tell whether colorama is enabled (analogous to the check a curses-based system would do by looking at the TERM environment variable). There doesn't appear to be any documented way to do this. We're considering a couple of undocumented hacks to discover this, but it would be better to have a documented and supported solution.

mivade:
Specifically, the hacks are to either check that `sys.stderr` is the same as `colorama.initialise.wrapped_stderr` or that `sys.stderr` is (on Windows) an instance of `colorama.ansitowin32.StreamWrapper`.

wiggin15:
The first suggested fix (check if sys.stderr is wrapped_stderr) won't work for sure because in some cases wrapped_stderr could be the original sys.stderr (not really a wrapped stream) - so they could be equal even if colorama is not in effect. This happens when `wrapper.should_wrap()` is False.

wiggin15:
What that means is that you can call `colorama.init` and it will not wrap anything. So maybe what we need is not `isinitialized`, but `is_wrapped`?
Something like this:

    def is_wrapped():
        return (isinstance(sys.stdout, StreamWrapper) or
                isinstance(sys.stderr, StreamWrapper))

bdarnell:
I think `is_wrapped` is too specific in light of #104 - in the future colorama might recognize that the terminal supports color without wrapping anything. What we want is basically "does stderr support color".

PR #201, opened 2018-10-23
Prevent user from wrapping already wrapped streams

tomchy:
Due to the fact, that colorama uses global streams, multiple wraps around the same stream is pointless and causes python to hung on both Linux and Windows-based machines.

If such behavior is allowed, huge delays inside modules, that uses colorama are introduced:
```
>>> import colorama
>>> import time
>>> for i in range(30):
...   start = time.time()
...   colorama.init()
...   stop = time.time()
...   print(stop - start)
...
0.002007007598876953
0.004018068313598633
0.003504037857055664
0.0030028820037841797
0.004488945007324219
0.005014896392822266
0.0025060176849365234
0.007011890411376953
0.00401616096496582
0.01253199577331543
0.013036966323852539
0.03408479690551758
0.03659701347351074
0.06968498229980469
0.13085198402404785
0.22510790824890137
0.6968679428100586
1.0372772216796875
1.97328519821167
3.6422548294067383
4.66950798034668
12.083894968032837
25.121819019317627
51.497376918792725
104.18486189842224
177.13460302352905
272.79527592658997
652.5946300029755
...
```

Related issues:
 - #128 
 - #140
 - #145
 - #182


Delgan:
What if user wants to `init()` colorama with `autoreset=True` while the stream has already been wrapped by a library with `autoreset=False`?
It looks like this implementation is going to prevent him from getting the desired result, right?

The hanging and huge delays while wrapping the same stream is a big issue indeed, it has been reported in #196 and I opened #198 to fix it.

Adding a safeguard arround multiple wrapping could be useful though to avoid `RecursionError` when `init()` is called thousands of times. However, if that happens, I think the user is doing something wrong.

Delgan:
Yeah, you are probably right and your example is pretty convincible. 🙂 

The fact that several people have reported encountering `RecursionError` problems is also a hint that something could be improved.

I can't think of any downside with your suggested change, but it should be interesting to know why it has not been implemented like this in the first time and why nested wrapping had been chosen instead.

The only surprising effect would be in this case:
1. Someone call `init()`
2. Someone wrap or replace `sys.stderr` with another stream-object
3. Someone call `init()` again

Then, the resulting `sys.stderr` would not be the one expected, but I don't know if such situation happens that often.

tartley:

tartley:
This PR dupes fixes in https://github.com/tartley/colorama/pull/236 and https://github.com/tartley/colorama/pull/262 (although this one predates those)

PR #210, opened 2019-01-05
Adding goto() and size() functions.

asweigart:
The goto() function makes it easy to reposition the cursor using 0-based coordinates as x, y (instead of the confusing y, x that curses uses). The size() function determines the size of the terminal window if possible (works on Windows, macOS, and Linux).

I think these are small additions to colorama that, while not color-specific, are likely to be used by folks using colorama as a cross-platform curses module.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -154,8 +154,6 @@ ANSI codes to reposition the cursor and get the size of the terminal window are
 supported.
 
     from colorama import init, goto, size
-
-    # use Colorama to make Termcolor work on Windows too
     init()
 
     goto(0, 0)  # move cursor to top-left corner
@@ -165,6 +163,13 @@ supported.
 
 See ``demos/demo06.py`` for an example of how to generate them with raw ANSI
 codes.
+
+The screen can be cleared of all text and the cursor reposition to 0, 0:
+
+    from colorama import init, clear
+    init()
+
+    clear()
 
 
 Init Keyword Args
--- a/README.rst
+++ b/README.rst
@@ -150,8 +150,21 @@ perform this reset automatically on program exit.
 Cursor Positioning
 ------------------
 
-ANSI codes to reposition the cursor are supported. See ``demos/demo06.py`` for
-an example of how to generate them.
+ANSI codes to reposition the cursor and get the size of the terminal window are
+supported.
+
+    from colorama import init, goto, size
+
+    # use Colorama to make Termcolor work on Windows too
+    init()
+
+    goto(0, 0)  # move cursor to top-left corner
+    goto(10, 0) # move cursor to x, y position of 10, 0
+    width, height = size() # get size of terminal window
+
+
+See ``demos/demo06.py`` for an example of how to generate them with raw ANSI
+codes.
 
 
 Init Keyword Args
--- a/colorama/__init__.py
+++ b/colorama/__init__.py
@@ -1,6 +1,6 @@
 # Copyright Jonathan Hartley 2013. BSD 3-Clause license, see LICENSE file.
 from .initialise import init, deinit, reinit, colorama_text
-from .ansi import Fore, Back, Style, Cursor, goto, size
+from .ansi import Fore, Back, Style, Cursor, goto, size, clear
 from .ansitowin32 import AnsiToWin32
 
 __version__ = '0.4.1'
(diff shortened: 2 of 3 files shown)


asweigart:
(Also, I have a simple clear() function that I can post a PR for if you want to accept this.)

asweigart:
My perspective comes from teaching beginning programming and usablity design. While the goto/size/clear/title behavior is currently *possible*, there's a large benefit that Colorama can capitalize on by making them *easy*. This doesn't just benefit new programmers; Python's gentle learning has made it adopted by beginners but also experienced developers.

I'm hesitant to cram dozens of features into a single module; I don't want to recreate curses. But the addition of the goto/size/clear/title functions would result in a large benefit for fairly little cost: The implementations are straightforward, unlikely to change, and easy to document.

I think I can make a justification for each of these changes. I don't want to add features just for the sake of adding features. "Simple is better than complex."

The current way to move the cursor is this:

sys.stdout.write('\x1b[%d;%dH' % (y + 1, x + 1))

This approach has a hard to remember control code, the y coordinate comes first, and is 1-based instead of the more typical 0-based coordinates. The sys.stdout.write part also reminds me of Java's verbose System.out.print That's a lot for a user to remember. We could instead replace it with this:

colorama.goto(x, y)

Technically, this isn't new functionality: colorama can already move the cursor by letting the user print this control code. But by repackaging this functionality into an easier API, it'll be used much more often.

The current way to clear the screen is this:

sys.stdout.write(colorama.ansi.clear_screen())

The function name is a bit confusing, because it implies that it clears the screen when it actually returns the string that, when printed, clears the screen. It makes a one-step process into a two-step process. Also, being in the ansi submodule, it's easy to mistakenly try to call colorama.clear_screen(). We could instead replace it with this:

colorama.clear()

The name is shorter and fits the "verb" use case, as the function itself clears the screen. Also, by being in the colorama namespace, we get the benefits of "flat is better than nested". We don't have to replace the clear_screen() function, so backwards compatibility is maintained.

The current way to set the terminal title is this:

sys.stdout.write(colorama.ansi.set_title(text))

This has the same issues as clear_screen(): it's inside a submodule and needs an additional step of being printed despite the "verb" style name.

The current way to get the terminal size is this:

import os
cols, lines = os.get_terminal_size()

This involves importing another module (which doesn't have any other terminal-based functionality; I could see users misremembering whether or not get_terminal_size() was in os or sys.) We could instead replace it with this:

cols, lines = colorama.size()

Not only is the name simple, but it's also grouped with the rest of the limited terminal functionality of colorama.

Again, these aren't completely unrelated to Colorama currently, as they are enabled and also have some functions already in ansi.py. This change makes these features easier to use. Python's lack of a cross-platform curses module is a huge opportunity for Colorama. I wouldn't want to press for features beyond the goto/size/clear/title ones in this PR, but for about a dozen lines of code we get a big leap in usefulness.

I didn't even realize there was a clear_screen() function in the ansi.py file. After adding these features, I'd like to go through and write up a guide that can be hosted on readthedocs.org so that all of Colorama's features get a better spotlight on them. Thanks for your work maintaining this module!

wiggin15:
I think you misunderstand the purpose of this module. colorama doesn't contain functions that manipulate the terminal. The only functions in colorama that are meant to be called directly help to generate ANSI codes - not "do things". This isn't a substitute or a subset of curses. colorama's only purpose is (according to the project description) to make "ANSI escape character sequences (for producing colored terminal text and cursor positioning) work under MS Windows." So colorama only contains code that translates codes that are _printed_, rather than directly doing what the codes are meant to do. If you are looking for a module that contains functions for directly manipulating the terminal, this isn't the place.

> sys.stdout.write('\x1b[%d;%dH' % (y + 1, x + 1))

You can simplify this code with `print(colorama.Cursor.POS(x, y))` (you can use `end=''` parameter to `print` if you want to avoid the trailing space). I already wrote this in my previous comment...

> The current way to get the terminal size is this:
> 
> import os
> cols, lines = os.get_terminal_size()
> 
> This involves importing another module (which doesn't have any other terminal-based functionality; I could see users misremembering whether or not get_terminal_size() was in os or sys.) 

This is not entirely correct. `get_terminal_size` is part of `shutil` (like I wrote before), not `os`, and shutil is relevant because it contains "shell utilities". I don't see the point in cloning functionality that is included in the Python standard library ("There should be one-- and preferably only one --obvious way to do it.").

> Not only is the name simple, but it's also grouped with the rest of the limited terminal functionality of colorama.

Again - this is important - colorama isn't about "terminal functionality" - it's about using and translating printed ANSI codes.

PR #216, opened 2019-03-16
Add some styles

SeaHOH:
1. UNDERLINE/UNDERLINE_OFF
1. OVERLINE/OVERLINE_OFF
1. REVERSE/INVERSE_OFF
1. CONCEAL/REVEAL

Only tested on Win7.


SeaHOH:
@wiggin15 Can you tell me what you think about this PR? Thanks.

wiggin15:
These new styles don't work on Windows terminals (cmd/PowerShell) and some don't have wide support in Unix terminals either (e.g. `OVERLINE` doesn't work in iTerm). I tried your branch on Windows 7 and couldn't see them.
Since there is no Windows support we decided not to support these styles (see https://github.com/tartley/colorama/issues/38#issuecomment-128834060)

SeaHOH:
Thanks for review.
How about `REVERSE/INVERSE_OFF` and `CONCEAL/REVEAL`? I think them works well.

Evidlo:
@wiggin15 A later comment in #38 mentions that underline, reverse and conceal are supported in modern powershell.

tartley:

JamesTheAwesomeDude:
I would appreciate support for the `'\x1b[8m'` "conceal" command, if that ever gets merged.

MATE terminal supports it \(haven't tested on other terminal emulators yet\), but on Windows CMD with Colorama 0.4.6, it doesn't work.

PR #221, opened 2019-04-20
[FIX] Fix example with ANSI sequence

aadibajpai:
`\033[30m` is the sequence for black and I was pretty worried because I assumed it would reset but my text just disappeared (I use white text on black bg so black text disappears)

`\033[39m` is the code for reset that should be used here.


wiggin15:
Thanks!

PR #222, opened 2019-05-25
Fix special ansi sequences not ignored and producing wrong colors

### LINKED ISSUE #217: 24 bits ANSI codes are wrongly converted instead of being stripped
Hi.

Some terminals support [24-bit ANSI colors](https://en.wikipedia.org/wiki/ANSI_escape_code#24-bit (aka "true colors") in the form `ESC[ 38;2;<r>;<g>;<b>`.

This code:

```python
import colorama
colorama.init()
print("\x1b[38;2;166;226;46m{}\x1b[0m".format("Foo"))
```

Result on Windows:

1552843793-screenshot

Result on a true colors terminal:

1552844028-screenshot

I guess there may be an issue within the `colorama` regex in charge of parsing color codes. On Windows, this should be stripped, right?

Delgan:
Hi.

This changes is intended to fix #217.

On Linux, special coloration sequences may be initialized with `38` (foreground) and `48` (bakground) ansi codes. [8-bit](https://en.wikipedia.org/wiki/ANSI_escape_code#8-bit) looks like ``\x1b[38;5;46m`` and [24-bit](https://en.wikipedia.org/wiki/ANSI_escape_code#24-bit) looks like ``\x1b[38;2;120;33;255m``. These colors are not supported by Windows terminal as far as I know (maybe PowerShell console supports it?), so they should be ignored by ``colorama``, as it is the case for others special styles. 

Notes about my implementation:
* I tried to optimize for the main use case which happens when there is no such special sequence. The overhead is just one check against ``skip`` truthy value most of the time.
* There is some edge case like ``\x1b[38;0;31m`` or ````\x1b[38;48;5;46m``. According to my observations on Linux, if the ansi code folowing ``38`` is not ``2`` nor ``5``, the sequence should be ignored.

Feel free to suggest improvements. 👍


pfmoore:
> hese colors are not supported by Windows terminal as far as I know

Recent versions of the Windows 10 terminal [do support them](https://devblogs.microsoft.com/commandline/24-bit-color-in-the-windows-console/)

Delgan:
Hey @tartley, I just updated the PR by adding unit tests (and slightly modifying the implementation to make it more readable).

Please let me know what you think of it!

PR #228, opened 2019-08-26
Add support for Python 3.7 and 3.8, drop EOL 3.4

hugovk:
Also test on Python 3.8 beta, due out in October.

* https://www.python.org/dev/peps/pep-0569/

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -71,7 +71,7 @@ Copyright Jonathan Hartley 2013. BSD 3-Clause license; see LICENSE file.
 Dependencies
 ============
 
-None, other than Python. Tested on Python 2.7, 3.4, 3.5 and 3.6.
+None, other than Python. Tested on Python 2.7 and 3.4+.
 
 Usage
 =====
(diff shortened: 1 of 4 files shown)


hugovk:
Updated, Python 3.8 is now officially out!

README.rst
```diff
@@ -71,7 +71,7 @@ Copyright Jonathan Hartley 2013. BSD 3-Clause license; see LICENSE file.
 Dependencies
 ============
 
-None, other than Python. Tested on Python 2.7, 3.4, 3.5 and 3.6.
+None, other than Python. Tested on Python 2.7 and 3.5+.
```
wiggin15: I think I prefer to keep the explicit list of tested versions. When future versions get released, we can't say that they are tested until they are.

hugovk: Updated in https://github.com/tartley/colorama/pull/228/commits/9f03cb6c237ac420e60858f20ee775224cb5e186.

wiggin15:
Thanks!

hugovk:
You're welcome!

PR #232, opened 2019-10-10
use getattr instead of try / catch

kaozdl:
use getattr instead of try / catching the isatty attribute of stream.
This avoids the overhead of capturing and sending the traceback.
also results in cleaner code.

### DIFF
--- a/colorama/ansitowin32.py
+++ b/colorama/ansitowin32.py
@@ -45,12 +45,8 @@ class StreamWrapper(object):
         if 'PYCHARM_HOSTED' in os.environ:
             if stream is not None and (stream is sys.__stdout__ or stream is sys.__stderr__):
                 return True
-        try:
-            stream_isatty = stream.isatty
-        except AttributeError:
-            return False
-        else:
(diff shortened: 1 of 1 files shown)


Delgan:
Hi @kaozdl.

Just to be noted: the `try / catch` was used on purpose to avoid some edge cases leading to an exploding number of recursive calls, see https://github.com/tartley/colorama/pull/197 and https://github.com/tartley/colorama/pull/198.

At that time, the `try / catch` replaced `return (hasattr(stream, 'isatty') and stream.isatty())`. I don't know if risks are the same using `getattr()`.

kaozdl:
@Delgan getattr returns the property with whatever it has and in the case that it does not have that property it returns false.
So it should be the same as the `try / catch` but without the branching and without the exception catching wich is expensive

SeaHOH:
But your code was still incorrect, the right way is like this:
`return getattr(stream, 'isatty', None) and stream.isatty()`

tartley:

PR #236, opened 2019-11-11
Guard against multiple calls to init().

tartley:
BEWARE, I haven't tested this on Windows (because I
don't have a Windows box.) It should definitely get
tested there before we merge.

This should no longer lose the original stdout/stderr,
so that we can restore them with a single call to deinit(),
while still preserving the behavior that multiple calls
to init() can override the passed values for autoreset,
convert, strip, wrap.

This fix is not intended to reset output to default
foreground/background colors after the call to
'deinit()' -  that was never intended behavior.

### DIFF
--- a/colorama/initialise.py
+++ b/colorama/initialise.py
@@ -5,9 +5,10 @@ import sys
 
 from .ansitowin32 import AnsiToWin32
 
+UNSET = object()
 
-orig_stdout = None
-orig_stderr = None
+orig_stdout = UNSET
+orig_stderr = UNSET
 
 wrapped_stdout = None
 wrapped_stderr = None
@@ -28,8 +29,13 @@ def init(autoreset=False, convert=None, strip=None, wrap=True):
     global wrapped_stdout, wrapped_stderr
     global orig_stdout, orig_stderr
 
-    orig_stdout = sys.stdout
-    orig_stderr = sys.stderr
+    # Prevent multiple calls from losing the original stdout
+    if (
+        orig_stdout is UNSET and
+        orig_stderr is UNSET
+    ):
+        orig_stdout = sys.stdout
+        orig_stderr = sys.stderr
 
     if sys.stdout is None:
         wrapped_stdout = None
@@ -49,10 +55,13 @@ def init(autoreset=False, convert=None, strip=None, wrap=True):
 
 
 def deinit():
-    if orig_stdout is not None:
+    global orig_stdout, orig_stderr
+    if orig_stdout is not UNSET:
         sys.stdout = orig_stdout
-    if orig_stderr is not None:
+        orig_stdout == UNSET
+    if orig_stderr is not UNSET:
         sys.stderr = orig_stderr
+        orig_stderr == UNSET
 
 
 @contextlib.contextmanager
@@ -78,3 +87,4 @@ def wrap_stream(stream, convert, strip, autoreset, wrap):
         if wrapper.should_wrap():
             stream = wrapper.stream
     return stream
+
--- a/colorama/tests/initialise_test.py
+++ b/colorama/tests/initialise_test.py
@@ -6,12 +6,11 @@ from unittest import TestCase, main
 from mock import patch
 
 from ..ansitowin32 import StreamWrapper
-from ..initialise import init
+from ..initialise import deinit, init
 from .utils import osname, redirected_output, replace_by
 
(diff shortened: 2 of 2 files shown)


colorama/initialise.py
```diff
@@ -49,10 +55,13 @@ def init(autoreset=False, convert=None, strip=None, wrap=True):
 
 
 def deinit():
-    if orig_stdout is not None:
+    global orig_stdout, orig_stderr
+    if orig_stdout is not UNSET:
         sys.stdout = orig_stdout
-    if orig_stderr is not None:
+        orig_stdout == UNSET
```
wiggin15: I think this statement has no effect (did you mean `=`?)

colorama/initialise.py
```diff
@@ -5,9 +5,10 @@
 
 from .ansitowin32 import AnsiToWin32
 
+UNSET = object()
 
-orig_stdout = None
-orig_stderr = None
+orig_stdout = UNSET
```
wiggin15: What is the difference between `UNSET` and `None`? Can't we use `None` as "unset"?

colorama/initialise.py
tartley: good catch, oopsie!

colorama/tests/initialise_test.py
```diff
@@ -75,13 +74,16 @@ def testInitWrapOffDoesntWrapOnWindows(self):
...
-        with osname("nt"):
+    def testInitTwiceCanUndoneWithDeinitOnce(self):
+        with osname('nt'):
+            self.assertNotWrapped()
             init()
+            self.assertWrapped()
             init()
             self.assertWrapped()
+            deinit()
+            self.assertNotWrapped()
```
wiggin15: Maybe we should add another "init" to make sure init->deinit->init works

colorama/initialise.py
tartley: I was worried that the real sys.stdout might be None (we have a check for that on line 34 / 40 in the diffs of this PR). If that were the case, then when we store sys.stdout in orig_stdout, then orig_stdout will be set to None, and be indistinguishable from the state before 'init' has been called.

colorama/initialise.py
tartley: and, amusingly, fixing that assignment breaks a bunch of other tests. Thinking...

colorama/tests/initialise_test.py
tartley: Yes, I have arrived at the same conclusion! :-) Doing...

colorama/initialise.py
tartley: Was a result of tests not cleaning up properly on tearDown. Fixed now.

colorama/initialise.py
pslacerda: I agree with @wiggin15. I was worried about stderr but seems that POSIX and Windows require both stdout and stderr.

wiggin15:
I confirmed this on Windows 7 with Python 3.5 (yes, it is old, but I think it should be enough).

One small concern I have is that passing different parameters to multiple init calls can still change the final effect. For example, if I call `init` once with `autoreset=True` and once with `autoreset=False`, with the current code there will be an auto-reset code sent after each "write" because we'll have two "AnsiToWin32" objects wrapping each other - one that resets and one that doesn't. With the new code, we lose the multiple wraps, causing only the latest parameters to take effect, so if the last call passes `autoreset=False`, we will not have an auto-reset. The same goes for the other parameters, I guess. I don't know if this is important enough (wrapping multiple times is not really a good thing), but it could cause backward-compatibility breaks.

malekcellier:
Please consider releasing a new version, this PR does it for me on Windows 10.
Thanks!

tartley:
Thanks for the thoughts. I'll do it any day now... :-)

malekcellier:
Thanks a lot for responding! In my current project I had to resort to some inelegant monkey-patching which I will be more than happy to discard :-)

bl-ue:
@tartley has the day come yet? :smile:  I'd like this feature too.  May it merged and a new release cut, please?

Kevinrp01:
This PR would be a big help for some of my use cases....I'd like to build on it a bit so that future calls to init can tweak the autoreset behavior (for use within a context manager). However, I see this one is still open, so curious what the thoughts are on if this will eventually get merged or is there a reason this was never merged and I should perhaps do a new one?

PR #242, opened 2019-12-06
Fix README's link to enterprise docs

tartley:
Also tweak Makefile build/upload instructions.

Awaiting review from wiggin so we can get me test.pypi.org access, to test this before we release it.


Makefile
```diff
@@ -15,13 +15,12 @@ clean: ## Remove build artifacts and .pyc files
...
 .PHONY: clean
 
-sdist: clean ## Build an sdist file
-	python setup.py sdist
+build: clean ## Build an sdist and wheel
+	python setup.py sdist bdist_wheel
 .PHONY: sdist
 
-upload: clean ## Upload an sdist file
-	python setup.py sdist upload
```
wiggin15: If I am not mistaken, `setup.py sdist upload` means both "build an sdist and upload it" (rather than "upload an existing sdist"). This means "upload" also built the distribution. Do we want to keep this behavior and make the "upload" command run the "build" command? Or keep it like this so we need to run both "make build && make upload"? I'm fine either way.

wiggin15:
I have to merge this because the tag is on the side branch rather than on master

tartley:
> the tag is on the side branch

I really was a klutz that day, huh? I might write out a "release checklist" just so that it's harder for me to make any mistakes next time.

Issue #248, opened 2020-01-30
Test case failures on aarch64 


ossdev07:
**Test Command**: I am adding support for aarch64 in Colorama and executing the tests as per the Travis-ci script, below is the command I am using:

```python -m unittest discover -p *_test.py```

**Test Failure**: Two test cases are failing while running Colorama on aarch64 platform, Please have a look:

- testInitDoesntWrapOnEmulatedWindows 

- testInitDoesntWrapOnNonWindows

**Test Log**: Below are the test logs from my local environment:

```
======================================================================
FAIL: testInitDoesntWrapOnEmulatedWindows (colorama.tests.initialise_test.InitTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.7.5/lib/python3.7/site-packages/mock/mock.py", line 1330, in patched
    return func(*args, **keywargs)
  File "/home/travis/build/ossdev07/colorama/colorama/tests/initialise_test.py", line 50, in testInitDoesntWrapOnEmulatedWindows
    self.assertNotWrapped()
  File "/home/travis/build/ossdev07/colorama/colorama/tests/initialise_test.py", line 35, in assertNotWrapped
    self.assertIs(sys.stdout, orig_stdout, 'stdout should not be wrapped')
AssertionError: <colorama.ansitowin32.StreamWrapper object at 0xffffb14dfdd0> is not <_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'> : stdout should not be wrapped
======================================================================
FAIL: testInitDoesntWrapOnNonWindows (colorama.tests.initialise_test.InitTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/ossdev07/colorama/colorama/tests/initialise_test.py", line 55, in testInitDoesntWrapOnNonWindows
    self.assertNotWrapped()
  File "/home/travis/build/ossdev07/colorama/colorama/tests/initialise_test.py", line 35, in assertNotWrapped
    self.assertIs(sys.stdout, orig_stdout, 'stdout should not be wrapped')
AssertionError: <colorama.ansitowin32.StreamWrapper object at 0xffffb14dfed0> is not <_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'> : stdout should not be wrapped
``` 

@wiggin15 As per the discussion on #209 I have raised this issue, Please share your thoughts on this.

ossdev07:
[...]

PR #253, opened 2020-03-20
Added arm64 jobs in Travis-CI

odidev:
Travis-CI has added support for ARM64. Added ARM64 jobs in Travis-CI for python v3.7.

As per the discussion https://github.com/tartley/colorama/issues/248

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,16 @@
 language: python
 cache: pip
-python:
-  - "2.7"
-  - "3.8"
-  - "3.7"
-  - "3.6"
-  - "3.5"
-  - "pypy"
+matrix:
+  include:
+    - python: 2.7
+    - python: 3.8
+    - arch: arm64
+      python: 3.7
+    - arch: amd64
+      python: 3.7
+    - python: 3.6
+    - python: 3.5
+    - python: pypy
 install:
   - "pip install mock"
 script: "python -m unittest discover -p *_test.py"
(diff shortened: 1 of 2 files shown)


odidev:
Skiped the two test cases for arm64 only when stdout on the platform is not a tty. Please do revert if any more input is required.

odidev:
Ping @wiggin15 Please could you have a look at this PR.

wiggin15:
Hi @odidev . The patch looks good. The only issue is about the other tests in `test_initialise.py` in the arm64 travis environment (and environments like it). The tests may pass, but because of the wrong `isatty` value, they may not really test what they should. Because of this, I think that if we add "skip"s for this case it would be more correct to put it on all the tests of this file. It would be simplest to put a `skipUnless` on `setUp` to achieve that.

odidev:
@wiggin15 As per your suggestion I have added ```skipUnless``` on ```setUp```, Please have a look at the logs: https://travis-ci.com/github/ossdev07/colorama/builds/158676030

Here are the changes: https://github.com/ossdev07/colorama/blob/master/colorama/tests/initialise_test.py

Please revert if this looks good to you for the merge, I will commit the same in the PR.

odidev:
@wiggin15 committed again with the changes you suggested, Please have a look and revert if something else is required to be done from my end.

wiggin15:
Thank you!

odidev:
@wiggin15 Thanks for accepting this PR.

PR #257, opened 2020-04-07
Support possibility to wrap either stdout or stderr

barabanus:
It's common to use `stdout` to generate binary files and `stderr` to print log and errors. At the same time `colorama` provides very convenient functionality to strip color codes from streams when output is redirected. The problem is that colorama wraps both `stdout` and `stderr` which strips binary symbols from `stdout` when binary files are generated.

It would be very useful for colorama library to wrap separately `stderr` / `stdout` on demand during initialization. I suggest to reuse 'wrap' argument to `colorama.init()` like this:
```
colorama.init()
colorama.init(wrap=False)
colorama.init(wrap="stdout")
colorama.init(wrap="stderr")
```

### DIFF
--- a/colorama/initialise.py
+++ b/colorama/initialise.py
@@ -22,23 +22,26 @@ def reset_all():
 
 def init(autoreset=False, convert=None, strip=None, wrap=True):
 
+    if wrap not in [True, False, "stdout", "stderr"]:
+        raise ValueError('wrap must be one of: True, False, stdout, stderr')
+
     if not wrap and any([autoreset, convert, strip]):
         raise ValueError('wrap=False conflicts with any other arg=True')
 
     global wrapped_stdout, wrapped_stderr
     global orig_stdout, orig_stderr
 
-    orig_stdout = sys.stdout
-    orig_stderr = sys.stderr
-
     if sys.stdout is None:
         wrapped_stdout = None
-    else:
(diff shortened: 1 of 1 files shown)


septatrix:
Why should this be restricted to `stdout` and `stderr`. Instead I would like passing the streams directly such that the signature would be `init(..., wrap: bool | TextIO) -> None` or similar. This would make it easier to use it with text buffers etc.

barabanus:
> I think I would prefer passing the streams directly instead of as strings such that the signature would be `init(..., wrap: bool | TextIO) -> None` instead.

The only valid values for the stream argument would be either `sys.stdout` or `sys.stderr`, so I don't see any benefits to use stream vs string.

colorama/initialise.py
```diff
@@ -22,23 +22,26 @@ def reset_all():
...
     global wrapped_stdout, wrapped_stderr
     global orig_stdout, orig_stderr
 
-    orig_stdout = sys.stdout
-    orig_stderr = sys.stderr
-
     if sys.stdout is None:
         wrapped_stdout = None
-    else:
+    elif wrap != "stderr":
```
tartley, line 36: Forgive my very overdue drive-by review, but does this do the wrong thing if wrap==False ?

PR #260, opened 2020-04-26
Use built-in unittest.mock when available

mgorny:
Python has a built-in mocking support since 3.3.  Use the built-in
module to avoid the dependency on external mock whenever possible,
and fall back to it on older versions of Python.


wiggin15:
Looks good! Thanks.

PR #262, opened 2020-05-27
Add initialization counter

### LINKED ISSUE #205: Handle multiple calls to init() and deinit() properly
Looking at this code: https://github.com/tartley/colorama/blob/master/colorama/initialise.py#L23

It looks like the code currently doesn't properly handle calling `init()` multiple times, I'd expect the `init()` function to set some internal state to know that it is already inited (maybe with a counter, in case nested `init()`/`deinit()` pairs should be supported?). Same with calling `deinit()`, it doesn't seem to always restore things properly.

Example code:

    import colorama
    colorama.init()
    colorama.init()
    colorama.deinit()

(of course, it might not be that obvious, but imagine two libraries want to initialize colorama in some situations)

Maybe it should also support nested `init()`/`deinit()` pairs:

    import colorama
    colorama.init()  # Initialized now
    colorama.init()  # Still initialized, maybe increase "refcount"/"initcount"
    colorama.deinit()  # Still initialized, because init was called twice
    colorama.deinit()  # Actually deinitialize, as the "refcount"/"initcount" will go to zero

Similar logic probably applies to `reinit()`, not sure.

Gaming32:
Resolves #205.

### DIFF
--- a/colorama/initialise.py
+++ b/colorama/initialise.py
@@ -5,6 +5,7 @@ import sys
 
 from .ansitowin32 import AnsiToWin32
 
+_init_counter = 0
 
 orig_stdout = None
 orig_stderr = None
@@ -20,7 +21,7 @@ def reset_all():
         AnsiToWin32(orig_stdout).reset_all()
 
 
-def init(autoreset=False, convert=None, strip=None, wrap=True):
+def _init(autoreset=False, convert=None, strip=None, wrap=True):
 
     if not wrap and any([autoreset, convert, strip]):
         raise ValueError('wrap=False conflicts with any other arg=True')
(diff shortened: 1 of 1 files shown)


tartley:

The above is cut-and-paste to everyone. To you especially: thank you, that is a rubbish bug, and I'm really happy to see a fix for it!

Erotemic:
Getting this fixed would help my use cases. Alternatively, if I had a way of knowing if colorama was ever initialized (even if it was deinitialized later), that would be enough for me. Currently I'm testing if `colorama.initialise.atexit_done` is Falsy, which works, but may not be the most stable test for this.

Gaming32:
@tartley Any updates on this?

tartley:
No updates currently, no.

But! For various reasons, I plan to give Colorama some love soon.

tartley:
An alternative approach to these problems has been suggested. Instead of modifying the poorly thought out interface and behavior of the original 'init' function, which is nontrivial and might be relied upon by existing programs, a proposal has been made by njsmith to add a new function, in parallel with init, which behaves more sanely, and hopefully will address your issues with it.

See njsmith's final comment [here](https://github.com/tartley/colorama/pull/139) ("I would like to sneak one more feature in...") before I merged that PR. The intention is to merge that relatively trivial PR, then make a release.

PR #264, opened 2020-06-22
Moved the metadata into setup.cfg

KOLANICH:


tartley:
I like the idea of this change. Can I ask if you have particular motivations for submitting it, other than the fun of switching to the new & much-improved syntax?

An important but presumably easily-fixable detail: The CI build failed on this branch, so it can't be accepted as-is.

I'm reading around the idea, and notice conversations like https://github.com/pypa/sampleproject/issues/56,
which point out that end-users with very old versions of setuptools will not be able to install Colorama if we use this new syntax. Apparently [setuptools 30.3.0](https://setuptools.readthedocs.io/en/latest/history.html#v30-3-0), from Dec 2016, introduced the ability to use setup.cfg like this.

This doesn't rule out making the change, I just want to have the conversation, so that we all understand as best we can.

Presumably there are very few users who are stuck on such an old version of setuptools. However, Colorama is a very widely-used transitive dependency. For example, is used by 'pip'. So it seems like a bad outcome if someone stuck on an old setuptools for reasons beyond their control was unable to use 'pip' because of a change we made. I have no idea whether that scenario is realistic. Thoughts welcome.

PR #267, opened 2020-07-07
Adds support for reverse video and underline on Windows 10.

### LINKED ISSUE #265: Colors STOPPED showing in Windows 10, latest update.
Colors STOPPED showing in Windows 10, latest update. Still work in Ubuntu. Same code.
This is my latest windows update https://support.microsoft.com/en-us/help/4549951/windows-10-update-kb4549951

Thank you!

jpwroberts:
Also prevents SGR codes 30-37 from appearing BRIGHT when the terminal is using a bright colour by default.

closes #265


jpwroberts:
Hi,
This is my first time contributing to an open source project, and I'm not sure about the process.  Is there anything further I should be doing?
I found a paragraph starting "After you're happy with the proposed changes, you can merge the pull request", in the documentation, but it seems strange that I'd have the necessary access to do that.  I'm reluctant to experiment on a production system.

wiggin15:
Hi @jpwroberts. Thanks for the pull request!
Regarding the process, your PR is ok and it needs to be reviewed and accepted, and then merged, by the project maintainers (which are @tartley and myself). You don't need to do any merges yourself.
Unfortunately, lately we had less time to go over contributions and review them, especially with recent global events. Hopefully we'll get the chance to review this soon.

jpwroberts:
Thanks for clarifying.  I understand, these are chaotic times.

tartley:

jtesta:
@wiggin15 Any estimate as to when this PR will be merged?  I maintain a project that would benefit from underlining support.  Thanks in advance!

README.rst
```diff
@@ -143,7 +143,27 @@ Available formatting constants are::
 
     Fore: BLACK, RED, GREEN, YELLOW, BLUE, MAGENTA, CYAN, WHITE, RESET.
     Back: BLACK, RED, GREEN, YELLOW, BLUE, MAGENTA, CYAN, WHITE, RESET.
-    Style: DIM, NORMAL, BRIGHT, RESET_ALL
+    Style: DIM, NORMAL, BRIGHT, BRIGHT_OFF, REVERSE, UNDERLINE, RESET_ALL
+
+``Style.REVERSE_OFF`` and ``Style.UNDERLINE_OFF`` are provided to allow intent
```
wiggin15, line 148: If I understand correctly, regular ANSI-compliant terminals do have a special code for `REVERSE_OFF` and `UNDERLINE_OFF`, but Windows doesn't have an attribute to "clear" these.
Does this mean that we'll have default behaviors on Windows/Linux? i.e. the following two calls
`print(colorama.Style.REVERSE_OFF + colorama.Back.GREEN + "off?" + colorama.Style.RESET_ALL)`
and
`print(colorama.Style.REVERSE_OFF + colorama.Style.REVERSE_OFF + colorama.Back.GREEN + "off?" + colorama.Style.RESET_ALL)`
will have "REVERSE" off in both on Linux, but will have "REVERSE" on in the first call on Windows?
Maybe we can call a different function, or use a special value in `WinStyle` to make sure `OFF` means `OFF`?

README.rst
```diff
@@ -143,7 +143,27 @@ Available formatting constants are::
...
+(and the same for UNDERLINE).
+
+It's important to realise that ``Style.UNDERLINE_OFF`` and ``Style.REVERSE_OFF``
+are not guaranteed to turn the relevant effect off.  If you've called
+``colorama.init`` with ``autoreset=False``, and left the terminal in underline
+mode, then the next time either ``Style.UNDERLINE`` or ``Style.UNDERLINE_OFF``
+is encountered, underlining will be turned off.
+
+Note that REVERSE and UNDERLINE require Windows 10, they don't have any effect
+on Windows 7 or 8.
```
wiggin15, line 161: Technically it's not the Windows version but the specific terminal (cmd) on these Windows versions.
(did we ever try these values on PowerShell on older versions? It's not important, anyway).

wiggin15 (changes requested):

wiggin15:
Thanks for nudging me, @jtesta .
This PR is pretty good. I left some comments that I think may need attention. There is also a small conflict with the latest code, so it needs to be rebased / updated. Once these issues are addressed I can merge the PR.

CHANGELOG.rst
```diff
@@ -1,3 +1,10 @@
+0.4.4
+  * Added support for REVERSE video and UNDERLINE.
+  * ANSI codes CSI [ 0-37 m no longer render as BRIGHT if the default terminal
+    foreground color is BRIGHT.
```
jpwroberts, line 4: I may have been a bit over-zealous there. Definitely agree about the last two not being in the changelog. The first two result in a change in behaviour, especially no longer getting a BRIGHT foreground colour with CSI [ 30-37 m. That should be documented somewhere, but I'm happy to remove it from the changelog.

README.rst
```diff
@@ -254,7 +274,14 @@ The only ANSI sequences that colorama converts into win32 calls are::
...
     ESC [ 1 m       # bright
     ESC [ 2 m       # dim (looks same as normal brightness)
+    ESC [ 4 m       # underline *1*
+    ESC [ 7 m       # reverse video *1*
+    ESC [ 21 m      # double underline or normal brightness
     ESC [ 22 m      # normal brightness
+    ESC [ 24 m      # underline off *1*
+    ESC [ 27 m      # reverse video off *1*
+
+    *1. Not supported on Windows 7 or 8.*
```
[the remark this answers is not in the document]
jpwroberts: Agreed

README.rst
jpwroberts: That's the behaviour I expected (and intended) but there seems to be a problem there. I'll work on that (after rebasing to the current version).
I think the Windows behaviour can be fixed by tracking the current state of REVERSE and UNDERLINE. I'll make sure OFF really means off (and ON doesn't mean OFF if the attribute is already set).

README.rst
jpwroberts: True. This text will be removed when I fix the problem with turning the attributes off in Windows.

CHANGELOG.rst
```diff
@@ -1,3 +1,10 @@
...
+  * The demo01.py file now demonstrates reverse video and underlining effects.
+  * winterm_test.py updated to assert on term._default_style rather than _style.
```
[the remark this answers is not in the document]
jpwroberts: I may have been a bit over-zealous there. Definitely agree about the last two not being in the changelog. The first two result in a change in behaviour, especially no longer getting a BRIGHT foreground colour with CSI [ 30-37 m. That should be documented somewhere, but I'm happy to remove it from the changelog.

wiggin15:
Thanks for updating the PR. If I understand correctly, this is a different implementation that before, and we don't use the Win32 calls any more?

This commit contains changes from the already-current master branch (looks like it's a merge commit) - is it supposed to look like this? It's harder to review and I don't know what merging it like this will mean. https://github.com/tartley/colorama/pull/267/commits/752c327260d171e4b7856f0d611583d61352ed2a
There are also still conflicts - can you try to rebase your branch on top of master instead of using merge?

colorama/ansitowin32.py
```diff
@@ -80,12 +88,41 @@ def __init__(self, wrapped, convert=None, strip=None, autoreset=False):
...
 
+        conversion_supported = True
         on_windows = os.name == 'nt'
         # We test if the WinAPI works, because even if we are on Windows
         # we may be using a terminal that doesn't support the WinAPI
         # (e.g. Cygwin Terminal). In this case it's up to the terminal
         # to support the ANSI codes.
-        conversion_supported = on_windows and winapi_test()
+        if on_windows and platform.release().startswith('10'):
+            # Windows 10 supports ANSI sequences using virtual terminal mode.
```
wiggin15, line 98: Please move to a separate function

colorama/ansitowin32.py
```diff
@@ -80,12 +88,41 @@ def __init__(self, wrapped, convert=None, strip=None, autoreset=False):
...
+            # However, this has been turned off by default from an early Windows 10 version.
+            # We turn virtual terminal support back on, then set
+            # `conversion_supported` to False. Behaviour will then be the same
+            # as for Linux - colorama doesn't use win32 calls.
+            from ctypes import wintypes, byref  # import wintypes now we know we're on Windows.
+            old_mode = wintypes.DWORD()
+
+            stdout_handle = windll.kernel32.GetStdHandle(STDOUT)
+            # If the next call failed, we may be running on Windows pre 10.0.10586.
```
wiggin15, line 107: If I understand correctly, this call isn't supposed to fail (`GetConsoleMode` existed in prior versions), but it will not return the `ENABLE_VIRTUAL_TERMINAL_PROCESSING` flag which only exists on later versions. The "set" call with the flag may fail, though. Please update the comments to reflect this.

colorama/ansitowin32.py
```diff
@@ -80,12 +88,41 @@ def __init__(self, wrapped, convert=None, strip=None, autoreset=False):
...
+            # Leave ANSI handling to colorama, there's no virtual terminal support available.
+            if windll.kernel32.GetConsoleMode(stdout_handle, byref(old_mode)):
+                if not (old_mode.value & ENABLE_VIRTUAL_TERMINAL_PROCESSING):
+                    if old_mode_value is None:
+                        # Don't set this again. __init__ gets called twice, for stdout and stderr.
+                        old_mode_value = old_mode.value
+                        print("Stored old_mode_value = " + str(old_mode_value))
+                    # Only enable if it's not already enabled
+                    print("enabling virtual terminal for " + str(self.stream))
+                    if windll.kernel32.SetConsoleMode(stdout_handle, old_mode_value | ENABLE_VIRTUAL_TERMINAL_PROCESSING):
```
wiggin15, line 115: I am a bit reluctant to do this here. Normally "init" itself doesn't change any console properties, and everything is controllable with the init flags. This call is now made implicitly and in any case.

colorama/ansitowin32.py
```diff
@@ -80,12 +88,41 @@ def __init__(self, wrapped, convert=None, strip=None, autoreset=False):
...
         self.stream = StreamWrapper(wrapped, self)
 
+        conversion_supported = True
         on_windows = os.name == 'nt'
         # We test if the WinAPI works, because even if we are on Windows
         # we may be using a terminal that doesn't support the WinAPI
         # (e.g. Cygwin Terminal). In this case it's up to the terminal
         # to support the ANSI codes.
-        conversion_supported = on_windows and winapi_test()
+        if on_windows and platform.release().startswith('10'):
```
wiggin15, line 97: The comment above this line mismatches the code now. We do need to call `winapi_test` before using the other WinAPI functions inside the "if".

colorama/ansitowin32.py
```diff
@@ -80,12 +88,41 @@ def __init__(self, wrapped, convert=None, strip=None, autoreset=False):
...
+            # Windows 10 supports ANSI sequences using virtual terminal mode.
+            # However, this has been turned off by default from an early Windows 10 version.
+            # We turn virtual terminal support back on, then set
+            # `conversion_supported` to False. Behaviour will then be the same
+            # as for Linux - colorama doesn't use win32 calls.
+            from ctypes import wintypes, byref  # import wintypes now we know we're on Windows.
```
wiggin15, line 103: Please move relevant code to win32.py, that's where we check the imports and define the ctypes functions.

wiggin15 (changes requested):

colorama/ansitowin32.py
jpwroberts: Actually, I think there's more than just the comment to change. The only way to be sure that we've enabled the virtual terminal is to call GetConsoleMode again, and check. Only if that test passes should I set conversion_supported to False.
To be safe, I need to find an old version of Windows 10 to test on. I'll also test on Cygwin.

colorama/ansitowin32.py
jpwroberts: After reading #300 I checked through the outstanding PRs. #139 by @segevfiner has used win32.py. That PR also enables vt via a function in winterm.py (as you suggested I should do). Once I've set up appropriate test environments (old Windows 10, Cygwin, etc), I'll run the same tests using #139. That may be a better candidate for implementing this functionality. @segevfiner identified vt as a more appropriate solution while I was still trying to make win32 calls behave.

colorama/ansitowin32.py
jpwroberts: I can't think of a cleaner way to do it, unfortunately. We're basically putting Windows into the same mode as Linux, and avoiding win32 calls. As far as I can see, that has to be done here.

jpwroberts:
> Thanks for updating the PR. If I understand correctly, this is a different implementation that before, and we don't use the Win32 calls any more?
> 
> This commit contains changes from the already-current master branch (looks like it's a merge commit) - is it supposed to look like this? It's harder to review and I don't know what merging it like this will mean. [752c327](https://github.com/tartley/colorama/commit/752c327260d171e4b7856f0d611583d61352ed2a)
> There are also still conflicts - can you try to rebase your branch on top of master instead of using merge?

It looks like I messed up there, sorry.

wiggin15:
After thinking about this further, I think the new approach is less than ideal, because it leaves older operating systems unsupported for specific new codes. colorama's purpose is to add support for terminals that don't support ANSI codes using Win32 API calls.

segevfiner:
This might mean you need both. As the Win 10 VT support enables support for additional escape sequences that aren't supported by the WinAPI such as 24-bit colors.

jpwroberts:
> After thinking about this further, I think the new approach is less than ideal, because it leaves older operating systems unsupported for specific new codes. colorama's purpose is to add support for terminals that don't support ANSI codes using Win32 API calls.

That's what I started off doing, then it became apparent that older Windows versions just don't have the necessary API calls to support the new codes I wanted to add (reverse video and underline).
This approach doesn't prevent adding new functionality to older operating systems - nothing has changed for them. We're just effectively treating Windows 10 the same as Linux.
I agree that we shouldn't do anything that disadvantages the older operating systems. If, for example, we discover a way to support strikethrough on Windows 7, that can be added to colorama in the usual way - regardless of this change.

jpwroberts:
It looks like I have a lot to learn about using github. I've just managed to accidentally close this PR while trying to delete all my previous commits.
I've made the changes suggested by @wiggin15 including moving appropriate code into functions in win32.py. But I'm really struggling to rebase this PR on top of master.
Unless someone is able to correctly rebase it, I think my best option is to open a new PR. I have local copies of all my proposed changes.

tartley:
Hi folks. I'm looking at Colorama stuff after many many months (years?) away. @jpwroberts, would you like some help getting this PR into shape? Do you think the basic approach is sound, and can solve (or already has solved) the issues @wiggin15 raised here?  Or have you given up on it? Thoughts appreciated.

tartley:
Hi. This PR has a laudable goal but cannot be merged in its current state (it contains diffs for many disparate changes and a large number of conflicts.) I'm closing it. Apologies for the extent to which my inattention has contributed to the difficulty of sorting this out. Hugs!

PR #276, opened 2020-10-12
Script to test releases

tartley:
Add a new 'test-release' script and makefile target

Running the script sends the already built package to
test pypi, then uses pip to download and install it from
there, and does a minimal invocation of the installed
package. If the package has not been built correctly,
this might detect some errors before we send it to
the real pypi.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ clean: ## Remove build artifacts, .pyc files, virtualenv
 
 $(virtualenv):
 	$(syspython) -m venv --clear $(virtualenv)
+	$(pip) install --upgrade pip
 
 venv: $(virtualenv) ## Create or clear a virtualenv
 .PHONY: venv
@@ -58,7 +59,7 @@ test-release: build
 	./test-release
 .PHONY: test-release
 
-upload: ## Upload our sdist and wheel
+release: ## Upload our sdist and wheel
 	$(twine) upload dist/colorama-$(version)-*
 .PHONY: release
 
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -48,11 +49,16 @@ test: ## Run tests
 
 # build packages
 
-build: clean ## Build an sdist and wheel
+build: ## Build an sdist and wheel
+	$(python) -m pip install --upgrade setuptools wheel
```
tartley, line 53: Independent of this PR's main point, doing 'clean' before attempting to build doesn't work. There is no $(python) to call, because the virtualenv has been cleaned.

In addition, we now update to latest setuptools and wheel before attempting to build wheels.

Makefile
```diff
@@ -48,11 +49,16 @@ test: ## Run tests
...
 	$(python) setup.py sdist bdist_wheel
 .PHONY: sdist
 
+test-release: build
+	./test-release
+.PHONY: test-release
+
 upload: ## Upload our sdist and wheel
-	$(twine) upload dist/*
+	$(twine) upload dist/colorama-$(version)-*
```
tartley, line 62: Independent of this PR's main point, we now just try to upload the current version, not everything in the dist dir.

tartley:
@wiggin15 FYI, this script to test new releases using the "test PyPI" passes when I run it from Linux. I'll try it on Windows too, then consider making a new release...

tartley:
I thought I'd be able to tweak a couple of things in the script to make it runnable from Windows, but am failing. I may replace the script with a .bat file...

PR #290, opened 2020-11-15
Add support and testing for PyPy3

jdufresne:


tartley:
This is fabulous. Thank you!

PR #292, opened 2020-11-22
Remove unused test imports and utils

jdufresne:
Removing unused imports exposed redirected_output() as an unused
function. It too has been removed.


tartley:
Brilliant, thank you very much!

PR #294, opened 2020-11-22
Use ASCII decoding explicitly in setup.py

### LINKED ISSUE #291: UnicodeDecodeError - 'ascii' codec can't decode characters in 0.4.4 README.rst
I was attempting to install a library that uses colorama in a Docker container and discovered that this was causing install failures from pip. I was using Python 3.6. I dug deeper and discovered that there are two "·" (dot) characters in the 0.4.4 README.rst that is being read as part of the setup.py. These characters cannot be decoded using 7 bit ASCII. They appear to no longer be present in the master branch, but a new release (or an update to 0.4.4) is likely necessary without these characters.

neirbowj:
Prevents future regression of #291 following 688f6cc

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@
 
 from __future__ import with_statement
 
+from io import open
 import os
 import re
 try:
@@ -14,8 +15,9 @@ except ImportError:
 NAME = 'colorama'
 
 
-def read_file(path):
-    with open(os.path.join(os.path.dirname(__file__), path)) as fp:
+def read_file(path, encoding='ascii'):
+    with open(os.path.join(os.path.dirname(__file__), path),
+              encoding=encoding) as fp:
         return fp.read()
 
 def _get_version_match(content):


neirbowj:
Dealt with a style nit.

yzgyyang:
Bump.

wiggin15:
As discussed in #291, README should be ascii and not unicode. Since README is now ascii again, a new release will solve this issue, so we don't really have to change the code.
We may want to set the encoding explicitly to "ascii" to avoid this mistake in the future - I will accept a PR for that.
(by the way, this is discussed in [PEP 597)](https://www.python.org/dev/peps/pep-0597/)

setup.py
```diff
@@ -3,6 +3,7 @@
 
 from __future__ import with_statement
 
+from io import open
```
wiggin15, line 6: Out of curiosity - is this for Python 2 compatibility?

neirbowj: Yes.

wiggin15:
Thank you!

PR #312, opened 2021-08-04
Added check to ensure init() will only have any effect in Windows.

coredumperror:
Much like #304, I am also experiencing issues because of the not-quite-true nature of that line from the docs that carltongibson quoted.

> On other platforms, calling init() has no effect

This is untrue when running in Linux within a Docker container, because even if you don't pass `strip=True` to `init()`, it'll still act like you did because in Docker, `sys.stderr` isn't a tty. It's a file descriptor that prints to a special log file which you can follow to get output that *looks* like a tty, but `isatty()` returns False. This forces `strip=True`, do to the `if strip is None` check in the `AnsiToWin32` constructor, which activates the `AnsiToWin32` wrapper even though the code isn't running on Windows. This completely breaks the colored output of all logs coming out of the Docker container.

This pull request makes it 100% certain that Colorama won't screw around with `sys.stderr` or `sys.stdout` unless the program is actually running on Windows. Libraries like StructLog already wrap their call to `colorama.init()` in such a check, so I figure that Colorama itself should be doing this.

### DIFF
--- a/colorama/initialise.py
+++ b/colorama/initialise.py
@@ -25,6 +25,12 @@ def init(autoreset=False, convert=None, strip=None, wrap=True):
     if not wrap and any([autoreset, convert, strip]):
         raise ValueError('wrap=False conflicts with any other arg=True')
 
+    # This function is only intended to have any effect when it's run on Windows.
+    # To ensure that unexpected side effects do not occur on any other platform,
+    # we return immediately if not running on Windows.
+    if sys.platform != 'win32':
+        return
+
     global wrapped_stdout, wrapped_stderr
     global orig_stdout, orig_stderr
 


coredumperror:
Maybe make this part of a major release, then? Compatibility-breaking changes are OK in major releases. Finally bring Colorama to 1.0, since the project is essentially finished anyway, and people will accept that they have to make a conscious decision to upgrade.

tartley:
Thanks once again. This change sounds sensible but needs some thought.

1. The tests don't pass with this change in place.
2. The change needs it's own automated tests. Omitting them is punting work onto future developers, who have no way of knowing that their changes have broken this MP's code.
3. Not your fault, but the codebase seems to detect Windows with a mixture of os.name in some places, and sys.platform in others. We should normalize on one or the other. That isn't the responsibility of this PR, but in order to write sane test for (2) above, one might want to sort it out.

tartley:
Thanks heaps for this. FYI, an alternative approach has been mentioned. Instead of modifying the behavior of the existing 'init()' call, upon which some people may be relying, the idea has been floated to add a new method alongside 'init', which does something like what you (and others) require.

For details, please see the final final comment my njsmith here, just before I merged this: https://github.com/tartley/colorama/pull/139.

tartley:
On reflection, to be explicit, I'm going to close this PR in favor of the approach suggested by njsmith, with the intention of getting a relatively trivial PR merged which can do this, and then making a release.

PR #326, opened 2021-10-20
replace tox.ini with pyproject.toml

eine:
See https://tox.wiki/en/latest/example/basic.html?highlight=pyproject#pyproject-toml-tox-legacy-ini

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,9 @@
+[tool.tox]
+legacy_tox_ini = """
 [tox]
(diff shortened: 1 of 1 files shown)


tartley:
Thanks for the contribution. I think this is sensible, but will presumably conflict with https://github.com/tartley/colorama/pull/338.

After 0.4.5 is released and I get back from vacation in July, let's try to get these both merged and make a 0.4.6 release.

PR #327, opened 2021-10-26
Add ITALIC, UNDERLINE, BLINK, REVERSE, STRIKE to AnsiStyle

sundarnagarajan:
I got these escape codes from [wikipedia ANSI escape code](https://en.wikipedia.org/wiki/ANSI_escape_code) and I have tested them

### DIFF
--- a/colorama/ansi.py
+++ b/colorama/ansi.py
@@ -93,6 +93,11 @@ class AnsiBack(AnsiCodes):
 class AnsiStyle(AnsiCodes):
     BRIGHT    = 1
     DIM       = 2
+    ITALIC    = 3
+    UNDERLINE = 4
+    BLINK     = 5
+    REVERSE   = 7
+    STRIKE    = 9
     NORMAL    = 22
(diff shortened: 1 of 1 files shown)


mxrch:
> Thanks for the pull request! We've had several issues and pull requests dealing with adding more ANSI codes, but our decision as of now is not to add codes that we can't support on older Windows terminals. colorama's main purpose is to "translate" ANSI codes on Windows terminals, and that can't be done for these ANSI styles. See [#38 (comment)](https://github.com/tartley/colorama/issues/38#issuecomment-128834060) and [#122 (comment)](https://github.com/tartley/colorama/issues/122#issuecomment-377104387) and

Let me question your decision.
I don't think we should deny improvements because earlier versions don't support it, besides, Windows Terminal is now installed by default.
Instead, we should write on the readme that older terminals are not supported, and start adding features.

ktdreyer:
Agreed, colorama is the best-maintained terminal coloring library. It would be great to have this support in this library.

PR #328, opened 2021-10-28
Atexit fix

3tilley:
A library we use throws this annoying error when it's called in under certain conditions. It's probably doing something wrong, but I think `colorama` should be more resilient to this particular bad state. I've attached the script that demonstrates the error. I've tried to write a test for this, but the best I can to do honestly test the functionality is to spin up a new Python in a subprocess that calls `colorama`,  this seems awkward and was difficult to make portable. If it's your desire I'll do it though.

```
from __future__ import print_function

import codecs
import sys

from colorama import init

init()
sys.stdout = codecs.getwriter("utf-8")(sys.stdout.detach())

print("foo", sys.stdout)
```

Before and after the fix:
image

image

Rejected test attempt:
```
    def testDetachAfterInitDoesntThrow(self):
        fd, name = tempfile.mkstemp(suffix=".py", text=True)
        sample_python = """
from __future__ import print_function

import codecs
import sys

from colorama import init

init()
sys.stdout = codecs.getwriter("utf-8")(sys.stdout.detach())

print("Hello", sys.stdout)
        """
        temp_script = Path(name)
        with temp_script.open(mode="w") as f:
            f.write(sample_python)
        output = subprocess.check_output(["python", str(temp_script)], universal_newlines=True)
        print(output)
        assert "buffer" not in output
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -163,6 +163,11 @@ Available formatting constants are::
 ``Style.RESET_ALL`` resets foreground, background, and brightness. Colorama will
 perform this reset automatically on program exit.
 
+These are fairly well supported, but not part of the standard::
+
+    Fore: LIGHTBLACK_EX, LIGHTRED_EX, LIGHTGREEN_EX, LIGHTYELLOW_EX, LIGHTBLUE_EX, LIGHTMAGENTA_EX, LIGHTCYAN_EX, LIGHTWHITE_EX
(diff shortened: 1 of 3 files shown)


3tilley:
Great idea, added. I tried something similar before but wasn't able to detach a StringIO - I'm clearly don't have your deep stream knowledge!

3tilley:
I think this is just waiting a CI run, @tartley do you mind kicking it off?

3tilley:
Thanks so much! I literally looked at this in an output today and thought back to this PR

PR #330, opened 2021-11-16
CI: Replace deprecated pypyX with pypy-X.Y

hugovk:
`pypyX` is deprecated and is not available in newer images:
https://github.com/actions/setup-python/issues/244#issuecomment-925966022

Instead explicitly specify the version:
https://github.com/actions/setup-python#specifying-a-pypy-version


wiggin15:
Thanks!

PR #338, opened 2022-03-27
Update package metadata

ofek:
## Background

Hello there! The Python packaging ecosystem has standardized on the interface for build backends ([PEP 517](https://www.python.org/dev/peps/pep-0517/)/[PEP 660](https://www.python.org/dev/peps/pep-0660/)) and the format for metadata declaration ([PEP 621](https://www.python.org/dev/peps/pep-0621/)/[PEP 631](https://www.python.org/dev/peps/pep-0631/)). As a result, the execution of `setup.py` files is now [heavily discouraged](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).

So, I'm spending my free time updating important projects so that they are modernized and set an example for others 😄

## Summary of changes

This implements [PEP 621](https://www.python.org/dev/peps/pep-0621/), obviating the need for `setup.py`, `setup.cfg`, and `MANIFEST.in`. The build backend `hatchling` (of which I am a maintainer) is quite stable and actively developed, the only reason the version is 0.x is because 1.0.0 will drop support for Python 2. It's also on the major distribution channels such as [Debian](https://salsa.debian.org/python-team/packages/hatchling), [Fedora](https://src.fedoraproject.org/rpms/python-hatchling), [Arch Linux](https://archlinux.org/packages/community/any/python-hatchling/), [conda-forge](https://anaconda.org/conda-forge/hatchling), [Nixpkgs](https://github.com/NixOS/nixpkgs/blob/staging/pkgs/development/python-modules/hatchling/default.nix), etc.

Other projects have done this such as [pipx](https://github.com/pypa/pipx/blob/2ecc668acf472ad6956cc682499c077c1130d17e/pyproject.toml), [tox](https://github.com/tox-dev/tox/blob/dbbcb37a816b0fbc72280ef26aeca7dece8b2858/pyproject.toml), all [Datadog Agent integrations](https://github.com/DataDog/integrations-core/pulls?q=is%3Apr+author%3Aofek+in%3Atitle+Add+pyproject.toml+file), etc.

In comparison to `setuptools`, Hatchling has reproducible builds, supports the modern [PEP 660](https://peps.python.org/pep-0660/) editable installation standard, has a more intuitive plugin system, and the codebase is an order of magnitude smaller thus making it more auditable and less prone to bugs (`setuptools` is based on a [deprecated](https://peps.python.org/pep-0632/) standard library module that it now must vendor).

## Notes

- Similar to https://github.com/tartley/colorama/pull/228, this also drops support for EOL Python 3.x versions since `hatchling` and other build backends like `setuptools` have already done so. This will not break anything because `pip` only looks for version candidates that satisfy the `requires-python` field. Here are stats for the curious: https://pypistats.org/packages/colorama
- The source distributions on PyPI are erroneously shipping a build artifact `*.egg-info` from `pip install -e .`; this is now fixed:
   c


ofek:
Just FYI Hatch was adopted by the Python Packaging Authority:

- https://github.com/pypa/hatch
- https://mail.python.org/archives/list/pypa-committers@python.org/thread/J2BL5XTTOCXUCELUE7L5P5HHJ7NXZGZ5/

tartley:
Hi. Thanks for the contribution! We're talking about making a release, and this seems like a laudable change. I'd like to get it in if I can. @ofek Presumably your opinion is that it is currently good to merge as-is? I'll check it out...

Makefile
```diff
@@ -51,15 +50,14 @@ test: ## Run tests
...
+	$(python) -m build
 .PHONY: build
 
 test-release: build ## Test a built release
 	./test-release
 .PHONY: test-release
 
 release: ## Upload a built release
-	$(twine) upload dist/colorama-$(version)*{.whl,.tar.gz}
+	$(twine) upload dist/colorama-*{.whl,.tar.gz}
```
tartley, line 62: Is it worth guarding against there being multiple matches for dist/*.whl, maybe by the 'build' step removing any old versions? This seems rampantly ad-hoc and racey and error-prone. Any other thoughts are welcome. (e.g. can we invoke anything here to get the specific version?)

ofek: No need IMO because PyPI does not allow uploads that would overwrite existing artifacts

tartley:
I see there are failures on 2.7 on Windows. I'll try to run a build on master as a sanity check.

ofek:
Fixed, this looks good to me now!

ofek:
When you drop py2 just remove the pin

tartley:
We're scrambling to find a little time to make a release before going on vacation. In order to keep things simple, I'm going to postpone merging this right now, but will merge it and make another release after my vacation, around July 5th.

Thanks heaps for the contribution, and all your work leading up to this point! :-)

ofek:
Happy to help! Have fun 🙂

wiggin15:
Hi @ofek . Can you explain the benefits of using hatchling vs. using `python -m build` once we have `pyproject.toml`?
Maybe these are just two different build backends and `build` are using a name that is too generic - but my knowledge about these recent developments is not up-to-date so this is a bit confusing.

ofek:
@wiggin15 Hey! See https://peps.python.org/pep-0517/

`build` is a _build frontend_ like `pip`. So `python -m build` and `pip wheel .` both invoke the configured _build backend_ which is what builds the sdist & wheel. Hatchling is a _build backend_ like setuptools.

Some benefits:

- easier to configure
- reproducible builds
- plugin ecosystem
- small, modern codebase which makes it easier to audit

wiggin15:
Thanks. I think I got it.

Makefile
```diff
@@ -51,15 +50,14 @@ test: ## Run tests
 # build packages
 
 build: ## Build a release (sdist and wheel)
-	$(python) -m pip install --upgrade setuptools wheel
-	$(python) setup.py sdist bdist_wheel
+	$(python) -m pip install --upgrade build
```
wiggin15, line 53: Do you think maybe "build" should be in `requirements-dev.txt` (which is installed during `make bootstrap` - see the release process in `README-hacking.md`) instead of being installed here?

pyproject.toml
```diff
@@ -0,0 +1,64 @@
...
+include = [
+    "/colorama",
+    "/demos",
+    "/CHANGELOG.rst",
+]
+
+[tool.hatch.build.targets.wheel]
+include = [
+    "/colorama/*",
+]
```
wiggin15, line 64: Why don't the wheel and sdist have the same files included?

wiggin15:
I'm ok with merging this now that we released v0.4.5. @ofek can you please rebase the PR to resolve the latest conflicts? Thanks.

Makefile
ofek: I think not since it's unrelated to testing; I think `twine` should also not be in there

pyproject.toml
ofek: Tests weren't being shipped in the wheel already (rightly so imo) https://pypi.org/project/colorama/#files :

w

ofek:
Done!

Makefile
wiggin15: `requirements-dev.txt` is not just for testing - we install these packages when running `make bootstrap` before running `make build`/`make release` as part of the build/release process.

OTOH now I see that we also run `pip install` for `setuptools` and `wheel` as part of `make build`...

Anyway, this is not blocking a merge.

Makefile
```diff
@@ -51,15 +50,14 @@ test: ## Run tests
...
+	$(python) -m build
 .PHONY: build
 
 test-release: build ## Test a built release
 	./test-release
 .PHONY: test-release
 
 release: ## Upload a built release
-	$(twine) upload dist/colorama-$(version)*
+	$(twine) upload dist/colorama-*{.whl,.tar.gz}
```
wiggin15, line 62: We encountered an issue in our last release where the globs using `{}` didn't work in the Makefile. We had to remove them.

tox.ini
```diff
@@ -1,5 +1,6 @@
 [tox]
-envlist = py27, py35, py36, py37, py38, py39, py310, pypy, pypy3
```
wiggin15, line 2: If we are removing py35 and py36 from the testing matrix we should reflect that in the README (we list tested versions under `Installation`)

wiggin15:
Sorry for piling on more comments each time, and thanks for working with us :)
I feel like we're almost there!

ofek:
Addressed!

wiggin15:
Thank you very much @ofek . This brings colorama to the modern age of Python packaging 🥳

tartley:

ofek:
Awesome!

PR #340, opened 2022-03-31
Enfoce reuse compliance

hexagonrecursion:
The https://reuse.software specification is the standard for providing machine-readable licensing and copyright information.

I believe:
* machine-readable metadata is a good thing.
* licensing and copyright matter
* automation is a good thing.

Therefore I propose we:
1. Make colorama compliant with the https://reuse.software/ specification
2. Enfoce compliance via a github action

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2013-2022 Jonathan Hartley & Arnon Yaari
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
 # These are supported funding model platforms
 
 # community_bridge: # Replace with a single Community Bridge project-name e.g., cloud-foundry
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2013-2022 Jonathan Hartley & Arnon Yaari
+#
+# SPDX-License-Identifier: BSD-3-Clause
+
 name: lint_python
 on: [pull_request, push]
 jobs:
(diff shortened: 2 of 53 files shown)


tartley:
Hi. Thanks for the submission!

I have some sympathy for the goals of automatic license identification, I realize there is a real problem that it aims to solve.

This MP replaces the existing human-readable copyright information in Python files. That was a convention we regretfully adopted a few years ago in order for Colorama to be used within Chrome. I do not have any insight in whether Colorama is still used in Chrome, and if it is, whether this MP would break whatever technical, legal or procedural requirements the Chrome team are currently relying on. We'd need to answer those questions before we could delete or replace those existing copyright headers.

In general though, I have historically been quite strongly opposed to placing copyright information in every source file, since in the majority of cases every single file in a project has the same copyright information, so this is needless and redundant, and the situation is unambiguously described by a common-sense interpretation of a single LICENSE file. Any legal or technical system that doesn't support this simple and widespread solution seems broken to me, and I'm reluctant to adopt it. I hear arguments that this causes problems when people separate a single file from the rest of the project, since it has no embedded copyright information. Trying to plug that hole seems pointless to me - people can still cut and paste any portion of a file, with or without the copyright header, so the hole is not pluggable.

If it turns out lots of people feel strongly about this and I'm in a minority here, I'll reconsider. If that were to happen, I'd look more favorably on a PR which uses REUSE's bulk licensing functionality, to declare the license for all files in a directory in a single place: https://reuse.software/faq/#bulk-license, rather than at the top of every file. Could we even get a way with a single such declaration, covering top level "*" ?

hexagonrecursion:
Thanks for your reply. I'll switch to bulk licensing and resubmit.

PR #345, opened 2022-06-14
Add arnon

tartley:
Adding Arnon to the list of PyPI metadata authors


wiggin15:
We should rebase this change on top of master because it is mixing changes from another PR

tartley:
This is very out-of-date and has conflicts. Easier to just apply the tiny diff over again on HEAD

PR #346, opened 2022-06-14
Release checklist includes '.candidate' releases

tartley:
I had intended that adding this might remove the need to do releases to
test.pypi.org, but that is fully automated, so running that command on
each release doen't seem like a burden.

### DIFF
--- a/README-hacking.md
+++ b/README-hacking.md
@@ -66,10 +66,11 @@ target.
 
 ## Release checklist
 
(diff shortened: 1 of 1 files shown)


wiggin15:
We should use "b" or "rc" postfixes instead of ".candidate" to comply with PEP 440 (which is respected by pip)
See https://peps.python.org/pep-0440/#pre-releases

PR #348, opened 2022-08-05
add shortcut class to build a color string

Dragon-GCS:
Add: build a color string by creating an instance

### DIFF
--- a/colorama/color.py
+++ b/colorama/color.py
@@ -1,0 +1,51 @@
+from .ansi import Style, Fore
+
+
+RESET = Style.RESET_ALL
+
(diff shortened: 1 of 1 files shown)


tartley:
Hi @Dragon-GCS. Thanks for the enthusiasm to submit this PR. With sincere apologies, I'm going to decline it though. The README links to README-hacking.md, for people hacking on the Colorama code, which states "Generally, we are not keen on new features. Even if they are small. [...] This is especially true if they are new ways to generate ANSI codes", for reasons which are described there.

Apologies for not being able to be more receptive.

PR #352, opened 2022-10-17
Add colorama.just_fix_windows_console()

njsmith:
As discussed in https://github.com/tartley/colorama/pull/139

Re: the bugs you collected:

> * [Guard against multiple calls to init(). #236](https://github.com/tartley/colorama/pull/236)  Guard against multiple calls to init()
> * [Prevent user from wrapping already wrapped streams #201](https://github.com/tartley/colorama/pull/201) Prevent user from wrapping already-wrapped streams

`just_fix_windows_console()` is a no-op if run repeatedly, or if run after someone else has called `init`.

> * [Support possibility to wrap either stdout or stderr #257](https://github.com/tartley/colorama/pull/257) Support possibility to wrap either stdout or stderr (for binary stdout)

`just_fix_windows_console` only touches streams that are pointing to a Windows console. So I guess in theory it would still interfere with your random binary if you try dumping it on your terminal, but (a) no-one expects that to do anything useful, (b) it interprets the random binary the same way a Unix console would, and people seem to live with that. So I'm not worried about it.

I've tested this manually (running `demos/demo01.py`) on Linux/Win7/Win10.

The big remaining issue is that I'm not sure how to test this, because it touches global state, and `init` also touches global state, and I'm not sure how to make all the tests run in a single process.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -55,7 +55,8 @@ This has the upshot of providing a simple cross-platform API for printing
 colored terminal text from Python, and has the happy side-effect that existing
 applications or libraries which use ANSI sequences to produce colored output on
 Linux or Macs can now also work on Windows, simply by calling
-``colorama.init()``.
+``colorama.just_fix_windows_console()`` (since v0.4.6) or ``colorama.init()``
+(all versions, but may have other side-effects – see below).
 
 An alternative approach is to install ``ansi.sys`` on Windows machines, which
 provides the same behaviour for all applications running in terminals. Colorama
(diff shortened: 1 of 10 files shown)


njsmith:
> The big remaining issue is that I'm not sure how to test this, because it touches global state, and init also touches global state, and I'm not sure how to make all the tests run in a single process.

Never mind, fixed this.

tartley:
This looks amazing, works for me, thank you @njsmith !

LqdBcnAtWork:
I'm unsure what's up with 0.4.6, but `Cursor` appears to be broken on win10. Sort of.

`Cursor.UP` seems to do nothing.
`Cursor.POS` appears to work.

Also, is there a better way to get cursor position than `win32.GetConsoleScreenBufferInfo().dwCursorPosition`?

njsmith:
@LqdBcnAtWork [Microsoft says it should work](https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences#cursor-positioning), and it seems to work for me:
image
Do you have a test case you can share?

LqdBcnAtWork:
I discovered _another_ quirk with `colorama.win32.GetConsoleScreenBufferInfo()`. While this is probably a bad place to write it, I'm unsure of where I should.

Anyway, in windows default command prompt, running `str(colorama.win32.GetConsoleScreenBufferInfo())` yields something like this: `'(9001,120,56,0,7,27,0,56,119,63,120)'`. Those first two numbers are the terminal height and width. However calling the same function in something like the fancy new Terminal app for win10, the result is _very_ different `'(52,209,51,0,7,0,0,51,208,52,209)'`. It gives the actual space for text, no matter how much you can scroll back by. Now this is an issue because the CMD reports the cursor position from it's zero, while the Terminal reports the cursor position within the screen, but both translate the ANSI calls to be relative to the screen.

So, in my case I have a function that'll print `n` lines, move back up, `n` lines, then save the cursor position for printing/reprinting a line. This is so I can specify the amount of lines I may or may not print (but not track) before scrolling would happen. That way I can return to a line and update it's contents without worrying about having to know _exactly_ how many lines have been printed. But due to inconsistent execution between dev and production I need to be able to handle things like the CMD and the Terminal.

Fortunately `colorama.win32.GetConsoleScreenBufferInfo().srWindow.Top` reports correctly for both, so I just get the `Y` value for the cursor, and subtract that `Top` value from it, and add 1. That way I always get an accurate value for the cursor's line.

Windows console, always causing issues because it's different.

tartley:

PR #354, opened 2022-10-18
Add UNDERLINE and NEGATIVE to ansi.py

LqdBcnAtWork:
add UNDERLINE, and NEGATIVE support to the `AnsiStyle` class,

add RGB methods to `AnsiBack` and `AnsiFore` classes

tested RGB functions using:
```python
import colorama
from colorama import Fore, Back, Style, Cursor

colorama.just_fix_windows_console()

for r in range(0,256,15):
  for g in range(0,256,15):
    b=0
    print(Fore.RGB(256-r,256-g,255)+Back.RGB(r,g,b)+f"({r:3} {g:3})", end="")
  print()

print(Style.RESET_ALL, end="")
```
Result on win10:
image

### DIFF
--- a/colorama/ansi.py
+++ b/colorama/ansi.py
@@ -28,7 +28,7 @@ class AnsiCodes(object):
         # Upon instantiation we define instance attributes, which are the same
         # as the class attributes but wrapped with the ANSI escape sequence
         for name in dir(self):
-            if not name.startswith('_'):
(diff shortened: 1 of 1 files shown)


LqdBcnAtWork:
I'm still learning GIT, closing this commit, going to pull request the _right_ one. aka `master`

PR #359, opened 2022-10-24
Update README.rst

LqdBcnAtWork:
- Removed references to `Style`, `Fore`, `Back`, and `Cursor`.
  - Not part of this projects main goal.
  - The demo's could be considered enough documentation for them. As they're not the primary purpose anyway.
- Removed references to updating `Style`, `Fore`, `Back`, and `Cursor`.
  - Looks messy,
- Removed mention of `autoreset` flag in `init()`
  - This feature seems to be beyond the scope of this project.
  - Perhaps a demo or separate RST/MD file could be created with the purpose of documenting ALL features of this package.
    - This would also be a better spot to discourage the idea of expanding `Fore`, `Back`, `Style` and `Cursor`.

This is in response to this comment:
https://github.com/tartley/colorama/pull/355#issuecomment-1288306104

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -147,20 +147,7 @@ with it, also for backwards compatibility.
 Colored Output
 ..............
 
-Cross-platform printing of colored text can then be done using Colorama's
-constant shorthand for ANSI escape sequences. These are deliberately
-rudimentary, see below.
-
-.. code-block:: python
-
-    from colorama import Fore, Back, Style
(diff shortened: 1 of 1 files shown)


tartley:
I don't think removing the documentation of features is a good idea. Just because they are not the package's primary purpose, that isn't a reason to leave them undocumented.

PR #369, opened 2023-02-08
RGB and Hex integrated

Alex-Jando:
I added two new commands: "from_rgb" and "from_hex". Both of them work on the colorama.ansi.AnsiFore and colorama.ansi.AnsiBack classes. It works as expected with no issues. Tested only however in windows terminal.


Greggles182:
This would be very useful @Alex-Jando

Greggles182 (approved):
These changes appear to work and would be very useful.

PR #371, opened 2023-02-17
Fix deprecated actions

dloez:
Fix deprecation warning "The `set-output` command is deprecated and will be disabled soon" on `jobs.test.steps.pip-cache` by redirecting the output of the echo command to `$GITHUB_ACTIONS` and by using the shell bash to avoid the command not working on Windows default shell.

At first, I was going to submit the PR by using the shell bash not only in that specific step but after thinking about it I only used it on that specific step as using the bash shell in the tox execution step could break everything.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,12 +35,12 @@ jobs:
       - name: Get pip cache dir
         id: pip-cache
         run: |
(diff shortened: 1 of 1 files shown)


Borda (approved):
yes, this is correct

cc: @hugovk

PR #377, opened 2023-04-17
python pep and language updates

marksmayo:
- sorted and tidied imports
 - cleaned whitespace
 - fixed indentations
 - removed inheritance of object
 - removed redundant exception types
 - added fstrings instead of format/% strings
 - replaced lambda with defs


njsmith:
Thanks for your interest in colorama! But unfortunately this project's in maintenance mode -- meaning we're only accepting important bug fixes, not any new features or cleanups. Our core functionality of making legacy Windows terminals work better is complete, and everything else is better served by other packages -- we recommend checking them out and contributing to them!

PR #386, opened 2024-04-01
improve AnsiStyle

zoumingzhe:


zoumingzhe:
@tartley Can you merge this patch？

zoumingzhe:
@tartley Can you merge this patch？

tartley:
Hi there. I am terribly sorry but Colorama is not seeking to enhance it's ability to generate ANSI codes, instead preferring to focus on its core functionality of converting existing ANSI codes in the stdout/err stream into win32 calls. This is described in detail on the readme and contributing docs - apologies that they are lengthy and this is easily overlooked.

Instead, we direct users who wish to generate ANSI codes, in the way your PR facilitates, to instead use a different library that is dedicated to that purpose, such as rich, blessings, termcolor, etc, all of which do a much better job than we could ever hope to. That way each library can focus on doing one thing well.

Greatest apologies -Jonathan.

PR #397, opened 2024-11-09
Fix just_fix_windows_console to use the StreamWrapper

rmccampbell:
The AnsiToWin32 class is not meant to be used as a file proxy directly, although it does define the write method, but it defines a `stream` member of type StreamWrapper which forwards/intercepts all the file operations as appropriate. In `init` which calls `wrap_stream` this underlying stream object is assigned to sys.stdout/stderr, but the `just_fix_windows_console` function instead assigns the AnsiToWin32 instance directly, which will break trying to use any other methods than `write` on the standard streams.

This change fixes the behavior of `just_fix_windows_console` to use the StreamWrapper object as in `init`.


rmccampbell:
I think properly testing this would require a pre-Windows 10 build of windows, so I have not actually tested it. IDK if colorama has some test workflow that would be able to actually verify it.

PR #401, opened 2025-02-14
Add PyPy support

James-E-A:
The Windows CMD still uses Codepage 437 for regular I/O by default.

However, PyPy [currently tries to treat it as UTF-8](https://github.com/pypy/pypy/issues/2999).

This pull request should fix that, making the `just_fix_windows_console` function *even better*.

Is this something you'd be interested in?


James-E-A:
I *think* this is ready to merge, would appreciate any feedback.

PR #403, opened 2025-03-14
Fix CI

tartley:
I'm seeing some CI failures when trying to merge a very old feature branch in #402. I'm going to fix CI separately in this PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
           - {python-version: "pypy-2.7", toxenv: "pypy"}
           - {python-version: "pypy-3.8", toxenv: "pypy3"}
           - {python-version: "2.7", toxenv: "py27"}
-          - {python-version: "3.7", toxenv: "py37"}
+          - {python-version: "3.7.17", toxenv: "py37"}
           - {python-version: "3.8", toxenv: "py38"}
           - {python-version: "3.9", toxenv: "py39"}
           - {python-version: "3.10", toxenv: "py310"}


tartley:
I'm not looking at this right now, but to remind myself when I come back to it:

Many test suites are failing because the requested (platform, Python version) combo is not available. I experimentally changed one of the failing Python versions from "3.7" to "3.7.14" (to match the actual version available). This seems to have stopped all python tests from being run at all. Is my change invalid syntax, which has stopped all the tests running? One would have hoped an error would result. More thrashing around required...

hugovk:
Looking at the CI run: https://github.com/tartley/colorama/actions/runs/13848387957?pr=403

```
Error when evaluating 'runs-on' for job 'test'. .github/workflows/test.yml (Line: 10, Col: 14): Unexpected value ''
```

That doesn't help much, but running https://github.com/rhysd/actionlint on the files reports that actions/checkout@v2, actions/setup-python@v2, actions/checkout@v3, actions/setup-python@v4 and actions/cache@v3 are "too old to run on GitHub Actions".

My PR https://github.com/tartley/colorama/pull/371 updates them.

I also recommend re-opening and merging https://github.com/tartley/colorama/pull/371, which fixes another deprecation which will be disabled soon.

.github/workflows/test.yml
```diff
@@ -18,7 +18,7 @@ jobs:
           - {python-version: "pypy-2.7", toxenv: "pypy"}
           - {python-version: "pypy-3.8", toxenv: "pypy3"}
           - {python-version: "2.7", toxenv: "py27"}
-          - {python-version: "3.7", toxenv: "py37"}
+          - {python-version: "3.7.17", toxenv: "py37"}
```
hugovk, line 21: Hmm, seems this is what causes `Unexpected value ''`, because we would also need to change `"3.7"` to `"3.7.17"` on line 14.

That fixes the syntax error and means the job will start, but then fails like:

```
  Version 3.7.17 was not found in the local cache
  Error: The version '3.7.17' with architecture 'x64' was not found for Ubuntu 24.04.
  The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

https://github.com/hugovk/colorama/actions/runs/16121812976/job/45489220430

Alternatively, change them both back to `"3.7"` and it fails like:
```
  Version ~3.7.0-0 was not found in the local cache
  Error: The version '3.7' with architecture 'x64' was not found for Ubuntu 24.04.
  The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

3.7 has been EOL for two years though, and I'm not personally interested in putting effort into it supporting, but I'm happy to help either drop support for it (my preference), or remove it from the CI matrix.

See https://github.com/hugovk/pypi-tools/blob/main/charts.md#colorama which shows there are very few installs for 3.7 and older.

tartley: Agreed, we could drop support on `main`, and when the next release happens we'll bump the major version.

Issue #300, opened 2021-02-10
Better colorama maintenance & support is needed


tartley:
This issue is primarily intended for myself (@tartley) & @wiggin15 to chat about how best to serve the colorama project going forward. But I thought I'd post it here so the discussion is public.

## Context

Many years ago, I (the initial author of Colorama) stopped using Windows. This left me with little motivation to continue work on Colorama. I appealed to contributors for help with maintenance, and @wiggin15 stepped up to provide masses of work on keeping the project working, by submitting fixes and improvements, and dealing with other's posted issues and PRs.

This worked really well for a long time, but eventually we have accumulated some long-standing issues and PRs that have languished without getting the attention they deserve. Turns out, (almost) unpaid volunteer open source work leads to burnout, as other projects and aspects of life eventually have to take priority. :-)

Representative of this, I got an email recently from [a github issue](https://github.com/tartley/colorama/pull/230), asking:

> Did you make any progress in resurrecting your release process?

I feel bad that there are PRs people have submitted, which are languishing unmerged, and wanted to chat about what's best to do about it. The following seem like options. Opinions or other ideas are welcome.

a) Continue as-is, with minimal maintenance. I don't feel this is fair on users or contributors.

b) Commit ourselves to spending more time helping get these PRs merged. I am not motivated to do this. How do you feel, @wiggin15?

c) Be more explicit (in the README, etc) that colorama is not a very active project, and people should not expect to get new PRs merged, especially for new functionality. This would then be very similar to option (a), but at least it gives people a heads-up on the situation before they work on submitting new PRs.

d) Extend and continue the search for others to help with maintenance. @wiggin15's perspective on this, especially, would be most welcome.

e) Explicitly tell people that we would welcome & support a fork, if it helps with (d).

hugovk:
If it helps the maintainers, there's an estimated $50/month available via Tidelift: https://tidelift.com/lifter/search/pypi/colorama

tartley:
The $50 is real. Since that arose, I experimented with raising other small revenue, for example by adding PayPal donate links to the README. Those efforts have so far generated $0.

wiggin15:
Hi @tartley .
I also noticed there are issues languishing and not getting attention in the recent past.
There are some issues that I am reluctant to attend to, either because (i) I am unsure how helpful they will be for all users, or (ii) I am afraid of big changes in a project that has to be very stable because of its wide use.
[...]

PR #409, opened 2025-07-09
Drop support for EOL Python 2.7-3.8

hugovk:
Follow on from https://github.com/tartley/colorama/pull/403#discussion_r2190489805

Python 2.7-3.8 are EOL:

image

https://devguide.python.org/versions/

2.7 and 3.7 are also no longer available on GitHub Actions causing the CI to fail. This PR also takes the opportunity to drop EOL 3.8.

This allows future improvements such as f-strings.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,13 +11,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["pypy-2.7", "pypy-3.8", "2.7", "3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["pypy-3.8", "3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
         os: [ubuntu-latest, macos-latest, windows-latest]
         include:
           # Add new helper variables to existing jobs
-          - {python-version: "pypy-2.7", toxenv: "pypy"}
           - {python-version: "pypy-3.8", toxenv: "pypy3"}
-          - {python-version: "2.7", toxenv: "py27"}
           - {python-version: "3.7", toxenv: "py37"}
           - {python-version: "3.8", toxenv: "py38"}
           - {python-version: "3.9", toxenv: "py39"}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,12 +11,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["pypy-3.8", "3.7", "3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["pypy-3.8", "3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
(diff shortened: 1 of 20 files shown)


tartley:
Thank you @hugovk !

tartley:
@hugovk If you happen to also look at the lint failures, be aware that my vague preferences are:

1. I'm not that keen on failing lint checks that run in the cloud and hence require creating an account elsewhere. A dev should be able to clone the repo and immediately run all the CI checks locally. I'd be happy to just delete any such failing checks.

2. I suspect a bunch of our existing lint checks and formatting could be replaced by ruff. Perhaps all of them?

I'm happy to chat about alternative ideas on any of the above, but those would be my vague preferences.

tartley:
@hugovk Also, huge thanks for the contributions, but be aware that colorama is sort of easing into a graceful retirement. I haven't used Windows in over a decade, at least, and the need for colorama it is much diminished these days (modern windows terminals all display ANSI codes just fine), so I mostly think that its current install base comes from projects who installed it long ago when it was more useful, and have never had incentive to remove it since.  Since that install base is so large and includes important projects, I'm super keen not to break them. So new features in Colorama have a very high bar to be accepted.

In particular, lots of people have been keen to add features extending the convenience helpers like Fore.RED, to help callers generate ASCI codes or the windows equivalents. However, in general I'm not keen to flesh out generation of all the possible ANSI codes in colorama. My philosophy is that colorama should do one thing well, and that one thing is converting ANSI codes into windows equivalents. If callers want convenience wrappers to help them generate the ANSI codes in the first place, then they should look at the MUCH more capable alternatives like rich, click, blessings, etc. I regret even including Fore.RED, etc, in colorama at all.

Best.  Jonathan.

tartley:
@hugovk UPDATE: A third preference for lint checks in CI: Changes should facilitate devs being able to run all CI checks locally, for example by the github actions running things like 'make lint'. Then it is guaranteed EASY for devs to run the checks locally, and to be sure that they results they see are the SAME results that will occur in CI.

hugovk:
> 1. I'm not that keen on failing lint checks that run in the cloud and hence require creating an account elsewhere. A dev should be able to clone the repo and immediately run all the CI checks locally. I'd be happy to just delete any such failing checks.

I agree. I've only seen `safety` in one or two other projects, and it mostly involved handling false positives, so think we should remove it to get the CI back to green and useful. Let's do this first.


> 2\. I suspect a bunch of our existing lint checks and formatting could be replaced by ruff. Perhaps all of them?

Yep, looks like many of them can be replaced. Let's look at this after 1.

> A third preference for lint checks in CI: Changes should facilitate devs being able to run all CI checks locally, for example by the github actions running things like 'make lint'. Then it is guaranteed EASY for devs to run the checks locally, and to be sure that they results they see are the SAME results that will occur in CI.

Also sounds good, can be done either before or after 2.

---

Re: graceful retirement, yeah, I saw https://github.com/tartley/colorama/issues/300. A safe "maintenance mode" without new features sounds reasonable and it's good to transparent about it and careful about compatibility 👍

PR #418, opened 2026-03-30
feat: honour NO_COLOR and FORCE_COLOR environment variables in init()

### LINKED ISSUE #415: Support NO_COLOR / FORCE_COLOR in Colorama
Users want environment variables to control color output. Today, colorama.init() ignores NO_COLOR and FORCE_COLOR, so tools cannot globally disable color or force it in CI without code changes. Implement env-aware behavior so NO_COLOR removes ANSI from writes to wrapped standard streams, and FORCE_COLOR enables color even on non-tty outputs. Token strings (Fore/Back/Style) must remain unchanged.

Proposed solution:
Add optional, backward-compatible env handling in init. Accept color_env={"auto","off"} (default "auto") and env (a Mapping; defaults to os.environ). Apply behavior via Colorama's wrapping of standard streams only; do not alter token constants. Precedence: explicit arguments override env-derived behavior. FORCE_COLOR wins over NO_COLOR when both are set. On Windows, FORCE_COLOR must enable wrapping and ANSI-to-Win32 conversion so colors are emitted reliably; on non-Windows, leave convert at its default unless needed. Invalid color_env raises ValueError; a non-mapping env raises TypeError. Keep the public API and formatting unchanged otherwise.

juliosuas:
closes #415.

tools that vendor colorama have no easy way to globally disable color. NO_COLOR (https://no-color.org) and FORCE_COLOR are the de-facto convention at this point, and it'd be nice if colorama picked them up out of the box.

added a small `_apply_color_env()` helper that reads the two env vars and tweaks `convert`/`strip` in `init()` before the stream wrapping runs. explicit kwargs still win — so any existing code that passes `convert=True` or `strip=False` behaves exactly like before.

precedence is:
1. explicit kwargs (caller always wins)
2. FORCE_COLOR (force on, even if not a tty)
3. NO_COLOR (force off, any non-empty value)
4. default tty detection

tests cover the combinations and the kwarg-override path.


juliosuas:
A bit more context on the implementation: the `_apply_color_env()` helper is intentionally separate from `init()` for testability — it takes `environ` as a parameter, making it easy to test all five scenarios without monkey-patching `os.environ`. The function returns `(convert, strip)` as a tuple so `init()` can simply merge the values before stream-wrapping. Let me know if the approach looks good or if you'd prefer it integrated differently.

juliosuas:
Closing this stale feature PR to reduce maintainer noise. NO_COLOR/FORCE_COLOR support is useful, but it should be re-proposed fresh if the project wants it.

PR #430, opened 2026-07-21
Exclude tests from wheel distributions

### LINKED ISSUE #405: Avoid bundling tests in wheels
*colorama* currently bundles the tests inside the wheel distributions. These usually are of no additional value outside of the source distributions and thus just bloat the distribution. For this reasons, I recommend excluding them from future wheels.

This matches the official recommendation for Python packaging as well (see https://packaging.python.org/en/latest/discussions/package-formats/#what-is-a-wheel):

> Wheels are meant to contain exactly what is to be installed, and nothing more. In particular, wheels should never include tests and documentation, while sdists commonly do.

juliosuas:
## Summary

- exclude `colorama/tests` from wheel distributions
- keep runtime modules and wheel metadata unchanged
- validate wheel contents from both Make and PowerShell release paths

## Verification

- wheel validator passes with no bundled test files
- `pytest`: 38 passed, 14 skipped
- `unittest`: 52 cases, 14 expected platform skips
- focused Black and Flake8 checks pass

Closes #405.


juliosuas:
Accidentally closed when the source fork was deleted during a profile cleanup. Same commits restored here: https://github.com/tartley/colorama/pull/433

